Issue #2, opened 2018-05-04
Keyboard shortcut registration only works on Mac keyboards


pbaize:
Mousetrap event is registered to the `command` key. Should switch to `mod` to be cross-platform.

PR #4, opened 2018-05-04
Shortcut fix

pbaize:
fixes the keyboard shortcut on windows #2


pbaize:
This might be worth publishing an update to npm for

PR #35, opened 2018-07-23
Add group/ungroup events to snap and dock

tomrobertsOF:
Added events when windows are grouped/ungrouped. 
 - Exports function `Layouts.addEventListener`
 - Takes and event type argument (currently 'joingroup' and 'leavegroup') and a callback
 - Added some code to the demo apps to log to console when added/removed from a group


src/SnapAndDock/Client/main.ts
```diff
@@ -25,13 +25,20 @@ const getId = (() => {
...
         client.dispatch('undock', getId());
         console.log('Window un-docked via keyboard shortcut');
     });
+
+    // Register servive listener for callbacks
+    client.register('joingroup', () => {
+        window.dispatchEvent(new Event('joingroup'));
+    });
+    client.register('leavegroup', () => {
+        window.dispatchEvent(new Event('leavegroup'));
```
pjbroadbent: These events should match existing naming conventions (ie: `join-group`, `leave-group`).

This is potentially a bit confusing, given that there's also the concept of groups in the core API. May need to go with a more descriptive name, or think about our terminology.

src/SnapAndDock/Client/main.ts
```diff
@@ -41,3 +48,20 @@ export const undock = exportClientFunction(clientP, (client: ServiceClient) => a
 export const deregister = exportClientFunction(clientP, (client: ServiceClient) => async (identity: ServiceIdentity = getId()) => {
                               await client.dispatch('deregister', identity);
                           }) as (identity?: ServiceIdentity) => Promise<void>;
+
+
+enum GroupEventType {
+    'joingroup' = 'joingroup',
+    'leavegroup' = 'leavegroup'
```
pjbroadbent: If making an enum, use standard enum naming conventions for the constants.

i.e: `JOIN_GROUP = 'joingroup',`.

src/SnapAndDock/Service/main.ts
```diff
@@ -11,20 +17,48 @@ async function init() {
...
+/**
+ * Sends a service message to the specified SnapWindow
+ * @param {GroupEventType} action The type of event being raised. The client will listen based on this value.
+ * @param {SnapWindow} window The target to which the message will be sent
+ * @param {fin.OpenFinServiceProvider} provider Provider object wrapping an instance of the openfin layouts service
+ */
+function sendWindowServiceMessage(action: GroupEventType, window: SnapWindow, provider: fin.OpenFinServiceProvider) {
+    console.log('Dispatching window message: ', action, 'to window: ', window.getIdentity());
+    provider.dispatch(window.getIdentity(), action, {});
+}
```
pjbroadbent: I think we should find another way of communicating with the client-side API. Having to pass all messages through this file may get a bit unmaintainable.

Don't have any suggestions off the top of my head, will need some thought/discussion. Can always break-out as a separate story.

src/SnapAndDock/Service/main.ts
```diff
@@ -11,20 +17,48 @@ async function init() {
...
-        // tslint:disable-next-line:no-any
-        (window as any).service.deregister(identity);
+        (window as ServiceWindow).service.deregister(identity);
+    });
+
+    // Register listeners for window added/removed signals
+    (window as ServiceWindow).service.onWindowAdded.add((group, window) => {
+        if (group.length < 2) {
+            return;
+        }
```
pjbroadbent: Can we remove this check?

Compile-time checks on the `emit` method and runtime checks when adding callbacks should ensure that the argument count will always match.

Want to avoid having this kind of boilerplate in every signal callback.

src/SnapAndDock/Service/main.ts
```diff
@@ -11,20 +17,48 @@ async function init() {
...
+ */
+function sendWindowServiceMessage(action: GroupEventType, window: SnapWindow, provider: fin.OpenFinServiceProvider) {
+    console.log('Dispatching window message: ', action, 'to window: ', window.getIdentity());
+    provider.dispatch(window.getIdentity(), action, {});
+}
+
+/**
+ * List of the valid grouping events that can be passed to the client.
+ */
+type GroupEventType = 'joingroup'|'leavegroup';
```
pjbroadbent: This is defined twice - once in the client and once in the service.

One of those should be exported, and re-used in the other component. Not sure which way round makes most sense... Will also need to think about build dependencies, and what exactly we want to expose to applications.

src/SnapAndDock/Client/main.ts
tomrobertsOF: How about: `join-snap-group`, `leave-snap-group`?

src/SnapAndDock/Service/main.ts
tomrobertsOF: This is the only place currently with a reference to the service provider, so in my mind it makes sense to have anything that needs to interact with the service directly happen here (or at least imported and invoked here). Let's discuss a bit offline.

src/SnapAndDock/Service/main.ts
tomrobertsOF: This is not a guard against bad arguments being passed but a way of avoiding double messages anytime a single window is docked/undocked. The process for a single window snapping to a group is:

1. remove from it's own group (fires the onWindowRemoved signal)
2. add to new group (fires the onWindowAdded signal)

so a window gets both events anytime they are snapped or unsnapped. I would think a user of the API would expect only one or the other. These checks ensure that if a window is snapped into a group, only the joined event is fired, and vice versa for removed.

We could probably accomplish the same thing by adding some checks in `SnapWindow.setGroup` or `SnapGroup.addWindow` , but would then also need to add additional signalling to send those messages up to the top-level provider interactions.

src/SnapAndDock/Service/main.ts
tomrobertsOF: I went with service exports and client imports for now. Won't be hard to change it around if we change our minds

src/SnapAndDock/Service/main.ts
pjbroadbent: Sorry, somehow read that as `arguments.length`...

I see. Makes sense.

src/SnapAndDock/Service/main.ts
pjbroadbent: Agreed, can see why you put it here - but I suspect that the amount of boilerplate required to get the necessary data into this location is going to get out of hand as we add additional API's and callbacks/events. Particularly when we also integrate workspaces and tabbing, as they will have their own set of events.

IMO, we should find some way of letting other classes send events back to the client directly. Beyond the scope of this story though, will add something to the backlog.

PR #36, opened 2018-07-25
Allow windows/applications to opt-out of snap&dock

tomrobertsOF:
New `deregister` service endpoint and client-api method to invoke it.
 - Completely removes the window from the service, making it effectively invisible
 - Other windows completely ignore this window when snapping.
 - Snapped windows will be unsnapped when deregister is called on them.
 - Currently no way to re-register a window other than closing and reopening.


resources/SnapDockDemo/frameless-window-triggered-deregistered.html
```diff
@@ -0,0 +1,72 @@
...
+        document.body.style.backgroundColor = colors[n - 1];
+        const h1 = document.createElement('h1');
+        h1.innerHTML = `Window ${n}`;
+        document.body.appendChild(h1);
+        const btn = document.createElement('button')
+        btn.innerText = 'Undock'
+        btn.onclick = () => window.Layouts.undock()
+        btn.style = '-webkit-app-region: no-drag'
+        document.body.appendChild(btn)
+</script>
```
pjbroadbent: Need to find a way to reduce the duplication here. I imagine we're going to end up with a lot of test cases that require some kind of slight difference in behaviour.

Something to consider as part of the general service structure/conventions work. But for the time being, could at least extract a `.css` file and do a bit of cleanup.

tomrobertsOF: That's a fair point. We could probably do something with search string parameters and keep it all together. For now i've moved the styling and duplicated stuff into `style.css` and `main.js` which makes it significantly cleaner

PR #47, opened 2018-08-10
fix npm module publishing and fix hardcoded version in client

bschwinn:
alter folder structure for npm module so that package.json's version can be used when client connects to provider.


MichaelMCoates:
Rebase please!

bschwinn:
this approach isn't 100%, continuing work/investigation, will submit another PR.

PR #53, opened 2018-08-14
SERVICE-196 - API for Creating a Tab Set from Multiple Windows

JohnathonLamOF:
Placed in TabUtilities, this will be moved later together with  Getblob function


src/provider/tabbing/TabUtilities.ts
```diff
@@ -136,6 +136,48 @@ export async function initializeTabbing(message: TabWindowOptions, uuid: string,
 	group.switchTab({ uuid, name });
 }
 
+/**
+ * Takes a tabblob and restores windows based on the blob
+ * @param tabBlob Restoration data
+ */
+export async function addMultiplWindowsToTab(tabBlob: TabBlob[]): Promise<void> {
```
ryanpetersonOF: Fix spelling error and/or change function name

src/provider/tabbing/TabUtilities.ts
```diff
@@ -136,6 +136,48 @@ export async function initializeTabbing(message: TabWindowOptions, uuid: string,
...
+                await group.window.alignPositionToApp(newTab.window);
+            }
+        });
+
+        group.window.finWindow.show();
+        group.switchTab({ uuid: blob.groupInfo.active.uuid, name: blob.groupInfo.active.uuid });
+    });
+}
+
+(window as any).multiwindow = addMultiplWindowsToTab;
```
ryanpetersonOF: No window attachment needed

src/client/types.ts
```diff
@@ -146,17 +146,22 @@ export interface TabbedEventPayload {
 	tabGroupID: string;
 }
 
+
+export interface Dimensions {
+    x: number;
+    y: number;
+    width: number;
+    setHeight: number;
+    appHeight: number;
```
pjbroadbent: So is `setHeight` the height of the tab strip, and `appHeight` the height of the application window below the strip?

With a name like `setHeight`, it's hard not to read 'set' as a verb.

src/provider/tabbing/TabUtilities.ts
pjbroadbent: Maybe this is what Ryan was referring to, but this creates _multiple_ tab groups right? It's more "add multiple windows to multiple tab groups"?..

Is this leaking too much of layouts into tabbing? Should this function only create a single tab group, and layouts calls it multiple times?

JohnathonLamOF: I think we discussed this yesterday and we will leave as is?

test/provider/tabbing/TabUtilities.unittest.ts
```diff
@@ -0,0 +1,70 @@
...
+            // Act
+            createTabGroupFromMultipleWindows(tabBlob, null);
+
+            // Assert
+            expect(console.error).toBeCalledWith(expectedErrorMessage);
+        });
+    });
+
+    //describe("Pass valid blob", () => {
+    //    it("should call switch tab to the active tab", () => {
```
pjbroadbent: Can you use `it.skip` rather than commenting-out? Just so this shows in the test runner output as something that needs fixing.

src/provider/tabbing/TabService.ts
```diff
@@ -8,10 +8,22 @@ import { TabAPIActionProcessor } from "./TabAPIActionProcessor";
 import { TabGroup } from "./TabGroup";
 import { ZIndexer } from "./ZIndexer";
 
+/**
+ * Interface for the tab service
+ */
+export interface ITabService {
```
pjbroadbent: Is this used? If it's not required for the mock (not sure what's happening with that, since it's commented-out) then IMO we should remove.

test/demo/tabbing/createTabFromMultipleWindows.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+    
+
+    // Assert
+    const win1Bounds: NormalizedBounds = await getBounds(win1);
+    const win2Bounds: NormalizedBounds = await getBounds(win2);
+    const tabGroupBounds: NormalizedBounds = await getBounds(newTabGroupWindow!);
+
+    // Window Bounds equality check
+    assert.true(win2Bounds.bottom == win1Bounds.bottom, `bottom are win2: ${win2Bounds.bottom} win1: ${win1Bounds.bottom}`);
```
pjbroadbent: Can you use `assert.is(win2Bounds.X, win1Bounds.X)` for these assertions?

If the assert fails you'll then get the expected and actual values within the error message automatically, without having to add custom messages on each assert.

test/runner.js
```diff
@@ -63,7 +63,7 @@ build()
     .catch(fail)
     //Had to restrict pattern to only include 'provider' as we now have a mix of ava and jest based tests.
     //Will need to port one to the other at some point - needs some discussion first.
-    .then(OF_PORT => run('ava --serial', ['build/test/provider/*.test.js'], { env: { OF_PORT } }))
+    .then(OF_PORT => run('ava --serial', ['build/test/test/demo/**/*.test.js'], { env: { OF_PORT } }))
```
pjbroadbent: Change this root to also run the existing tests - `build/test/**/*.test.js`

test/demo/tabbing/createTabFromMultipleWindows.test.ts
```diff
@@ -0,0 +1,114 @@
...
+            { uuid: app2.identity.uuid, name: win2.identity.name! },
+        ]
+    }];
+
+    // Get the service window in order to be able to find the tabgroup window
+    const serviceApplication: Application = await fin.Application.wrap({ uuid: "Layout-Manager", name: "Layout-Manager" });    
+
+
+    // Act
+    const scriptToExecute: string = `createTabGroupsFromMultipleWindows(${JSON.stringify(tabBlobs)}).then(() => { fin.desktop.InterApplicationBus.send("TEST", "replytest", "script evaluated", () => { console.log('successfully published') }, console.error)})`;
```
pjbroadbent: Everything from the `.then` onward should be within the util/in the provider.

PR #55, opened 2018-08-15
[SERVICE-183] Add a registered function to Tabbing that returns the JSON needed to restore tabs.

ryanpetersonOF:
Allows save and restore to import a function and use it to retrieve tabbing information in the shape of a TabBlob[]:

```
interface TabBlob {
    groupInfo: {
        url: string;
        active: TabIdentifier;
        dimensions: {
            x: number;
            y: number;
            width: number;
            tabGroupHeight: number;
            appHeight: number;
        };
    };
    tabs: TabIdentifier[];
}

interface TabIdentifier {
    name: string;
    uuid: string;
}
```


src/provider/tabbing/SaveAndRestoreAPI.ts
```diff
@@ -0,0 +1,52 @@
...
+import { TabGroup } from "./TabGroup";
+import { TabService } from "./TabService";
+
+/**
+ * Gathers information from tab sets and their tabs, and returns as a JSON object back to the requesting application/window.
+ * @param uuid Uuid of the requesting Application
+ * @param name Name of the requesting window
+ * @returns {TabBlob[] | undefined} Returns undefined if a tab service is not around.
+ */
+export async function getTabSaveInfo(): Promise<TabBlob[] | undefined> {
```
pjbroadbent: This is a rather general point, and it may just be that I need to read through the tabbing code a bit more - but it's not really clear to me what the rationale is behind choosing if something is implemented as a function within `TabService`, as a stand-alone util function (like this), or as a method within a `*Processor` class.

src/provider/tabbing/SaveAndRestoreAPI.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	return Promise.all(
+		TabService.INSTANCE.tabGroups.map(async (group: TabGroup) => {
+			const tabs: TabIdentifier[] = group.tabs.map((tab: Tab) => {
+				return tab.ID;
+			});
+
+			const [groupBounds, appBounds] = await Promise.all([group.window.getWindowBounds(), group.activeTab.window.getWindowBounds()]);
+
+			const groupInfo = {
```
pjbroadbent: Can you run the linter on this? It's tab-indented, whilst the rest of the project is using spaces.

test/provider/TabbingSaveAndRestore.test.ts
```diff
@@ -0,0 +1,23 @@
...
+//TODO need mock fin API (??)
+
+describe("Tests for save and restore API methods", () => {
+    describe("Tests for getting tabbing info blob", () => {
+        describe("Request to get tab info with tabbing no service", () => {
+            it("should return undefined", async () => {
+                const tabBlob = await getTabSaveInfo();
+
+                expect(tabBlob).toBeUndefined();
+            });
```
pjbroadbent: Needs more coverage :)

I get this is likely blocked due to other discussions right now, though.

src/provider/tabbing/SaveAndRestoreAPI.ts
ryanpetersonOF: Processor usually handles anything that has an API endpoint.  Theres been discussion if some (most) of the functions in the utilities file should be moved under the tab service which we agreed needs to be done, we just need to actually do it. 

save and restore API was originally a class but since it was essentially just a function we decided to just export that instead.

test/provider/TabbingSaveAndRestore.test.ts
pjbroadbent: Additional testing will be added at a later date.

PR #56, opened 2018-08-15
[SERVICE-218] Refactor AddTab to Group

ryanpetersonOF:
No longer need to make additional calls to add a tab into a tab group.  The addTab method will handle everything, with optional params to not handle certain things as is rarely, but sometimes, required.


tomrobertsOF:
Other than the conflicts, this looks fine. Get those resolved and i'll approve

ryanpetersonOF:
@tomrobertsOF Ive fixed conflicts

PR #58, opened 2018-08-16
[SERVICE-179] Allow reordering of tab handles in tabstrip UI

tomrobertsOF:
Recreating after resolving merge conflicts. 

- Uses the Sortable.js package to handle drag-and-drop-reordering.
- Existing drag-and-drop functionality is preserved (tearouts and moving windows between tabsets)
- Skeleton code in place to publish order updates to the tabbing service (will call API method from SERVICE-216)


src/provider/tabbing/DragWindowManager.ts
```diff
@@ -56,7 +62,7 @@ export class DragWindowManager extends AsyncWindow {
                     opacity: 0.01,
                     frame: false,
                     waitForPageLoad: false,
-                    alwaysOnTop: true,
+                    alwaysOnTop: false,
```
pjbroadbent: I thought this needed to be on top? So that it captured all the mouse events?..

tomrobertsOF: Sooo, kind of. When experimenting with drag/drop reordering I found that it would probably be best to have it not quite on top, but rather it should be second in the zindex with the focused window on top.

The startDrag listener brings the invisible window to front, then the source of the event is pulled on top. This lets html5 drag work inside the window, while preserving the nice cursor when dragging outside.

tomrobertsOF:
I think it's best to leave this unmerged until SERVICE-216 is finished. Then we can complete this story with a single merge.

PR #59, opened 2018-08-16
[SERVICE-216] Add API method to update tab order from tabstrip

ryanpetersonOF:
Adds API to the tab strip API to let the service know when the order of tabs has been changed.  This is done by sending an array of {uuid, name} in the order that the tabs are in the strip from first (0) to last (n).

No tests


src/client/TabbingApi.ts
```diff
@@ -19,23 +19,48 @@ export class TabbingApi extends Api {
...
 	/**
 	 * @constructor
 	 * @description Constructor for the TabbingApi class
 	 */
 	constructor() {
-        super();
-        this.mWindowActions = new TabbingApiWindowActions();
+		super();
+		this.mWindowActions = new TabbingApiWindowActions();
+	}
```
pjbroadbent: Can you run the linter on this? It's tab-indented, whilst the rest of the app is using spaces.

src/provider/tabbing/Tab.ts
```diff
@@ -122,6 +128,21 @@ export class Tab {
...
+     * @returns {number} Order Index
+     */
+    public get orderIndex(): number {
+        return this._windowOrderIndex;
+    }
+
+    /**
+     * Sets the index for this tab in the group.
+     */
+    public set orderIndex(order: number) {
```
pjbroadbent: What ensures that this stays in-sync with the actual array of tabs?

Would having the array itself be the "source of truth" for tab ordering be a safer way to manage this?

src/provider/tabbing/TabGroup.ts
```diff
@@ -80,6 +80,31 @@ export class TabGroup {
...
     }
 
+    /**
+     * Reorders the tab structure to match what is present in the UI.
+     * @param {TabIdentifier[]} orderReference The order which we should rearrange our tabs to match.  This will come from the UI component.
+     */
+    public reOrderTabArray(orderReference: TabIdentifier[]): void {
+        if (this._tabs.length !== orderReference.length) {
+            console.error('Mismatched array lengths on reorder!');
+            return;
```
pjbroadbent: If this can exit without doing anything, there should really be some way for the callee to know it's changes weren't applied. An exception, or returning a boolean, or something. Same happens in the client side of this call - can fail without any way for the calling code to check.

src/provider/tabbing/TabGroup.ts
```diff
@@ -117,6 +142,10 @@ export class TabGroup {
 
         await tab.remove(closeApp);
 
+        this._tabs.forEach((tab: Tab, i) => {
+            tab.orderIndex = i;
+        });
```
pjbroadbent: Do we need something similar on `addTab`? Again, would make more sense to me if the array was the thing that defined the order of tabs.

ryanpetersonOF: AddTab will only ever place a tab on the end.  Therefore its safe to assume that when you call addTab, its orderIndex will be _tabs.length + 1.

src/provider/tabbing/TabGroup.ts
ryanpetersonOF: ClientSide has no way of knowing what data the service has on it so its hard to make this comparison there.  I agree that client side should get some feedback.  Ill look into how to achieve this.

PR #60, opened 2018-08-16
[SERVICE-216] Add API method to update tab order from tabstrip

ryanpetersonOF:
Adds Tab API method to send new tab orders from UI to the service.  You pass in an array to the method of the new order.

The source of truth is the service side array, using first index as first tab in the UI.


src/client/TabbingApi.ts
```diff
@@ -19,23 +19,48 @@ export class TabbingApi extends Api {
...
-        this.mWindowActions = new TabbingApiWindowActions();
+		super();
+		this.mWindowActions = new TabbingApiWindowActions();
+	}
+
+	/**
+	 * Sends a new tab order to the service when tabs are reordered in the UI.
+	 * @param {Tab[]} tabs Tab Identifiers in index order of first (0) to last (n).  
+	 */
+	public sendTabOrder(tabs: TabIdentifier[] = []) {
```
pjbroadbent: Can you remove the default value here? Seems strange to have it default to a value that will result in an error - better to have TypeScript ensure that you actually pass a value for the argument.

src/client/TabbingApi.ts
```diff
@@ -165,20 +190,20 @@ export class TabbingApi extends Api {
...
+		}
+		if (!uuid) {
+			console.error('No uuid has been passed in');
+			return;
+		}
+
+		if (!name) {
+			console.error('No name has been passed in');
+			return;
+		}
```
pjbroadbent: Can you run the linter on this? This is tab-indented, whilst the rest of the project is using spaces.

src/provider/tabbing/TabGroup.ts
```diff
@@ -80,6 +80,46 @@ export class TabGroup {
...
+        const newlyOrdered: Tab[] = [];
+
+        // Copy of the tabs we know of
+        const existingTabsCheck: Tab[] = this._tabs.slice();
+
+        orderReference.forEach((ref) => {
+            this._tabs.find((tab, i) => {
+                if (tab.ID.name === ref.name && tab.ID.uuid === ref.uuid) {
+                    newlyOrdered.push(tab);
+                    delete existingTabsCheck[i];
```
pjbroadbent: If I'm reading this right, you're expecting the index of a tab within `_tabs` to match the index within `existingTabsCheck`? This will be the case at first, but won't necessarily hold once you start removing items from `existingTabsCheck`.

Iterating over `_tabs` in reverse order should work, or looking-up the index within `existingTabsCheck` before you do the delete.

src/provider/tabbing/TabGroup.ts
```diff
@@ -80,6 +80,46 @@ export class TabGroup {
...
+        });
+
+        // Any tabs we know of that werent in the array to match against will be stuck on the end of the new order.
+        // Prevents user from putting nonexistant tabs to match against
+        newlyOrdered.push(...existingTabsCheck.filter((tab) => {
+            if (tab) return true;
+            return false;
+        }));
+
+        this._tabs = newlyOrdered;
```
pjbroadbent: I'm wondering if you could collapse this down a bit?..

```
const newlyOrdered = orderReference.map((ref) => this._tabs.find( ... )).filter(tab => tab !== undefined);

if (newlyOrdered.length === this._tabs.length) {
    this._tabs = newlyOrdered;
    return true;
} else {
    console.error("Input array must reference each tab exactly once");
    return false;
}
```

src/provider/tabbing/TabAPIActionProcessor.ts
```diff
@@ -78,6 +78,9 @@ export class TabAPIActionProcessor {
             case TabAPIActions.ENDDRAG:
                 this._endDrag(message as TabAPIDragMessage, tabGroup);
                 break;
+            case TabAPIActions.TABSREORDERED:
+                this._tabsReordered(message as {action: TabAPIActions, properties: TabIdentifier[]}, tabGroup);
```
pjbroadbent: Define an interface for this message type, same as the others?

src/provider/tabbing/TabAPIActionProcessor.ts
```diff
@@ -93,6 +96,15 @@ export class TabAPIActionProcessor {
         this.mTabService.dragWindowManager.show();
     }
 
+    /**
+     * Handles when the UI component reorders its tabs.  Needed for save and restore.
+     * @param message The array of tabs to sort by.
+     * @param group The TabGroup attached to this call.
+     */
+    private async _tabsReordered(message: {action: TabAPIActions, properties: TabIdentifier[]}, group: TabGroup) {
```
pjbroadbent: Not actually async?.. :-)

PR #61, opened 2018-08-20
[SERVICE-194] Tab Group Creation on Window Drag Over

tomrobertsOF:
Added ability to create tabgroup by dragging and dropping an untabbed window onto a tabbed window.


tomrobertsOF:
Waiting on feedback from Tabbing team about some high-level decisions. This may need some tweaking depending on the result of those.

PR #62, opened 2018-08-21
S&R Improvements/Bug Fixes

MichaelMCoates:
- Properly saves and restores minimized windows (including child windows)
- Speed up
- Placeholder window bug fix (was only showing one placeholder window for child windows)
- Removes manifest field from LayoutApp


src/client/types.ts
```diff
@@ -27,7 +27,7 @@ export interface WindowState extends Bounds {
 
 export interface LayoutApp {
 	manifestUrl?: string;
-	manifest?: any;
+	createdFromManifest: boolean;
```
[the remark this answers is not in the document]
MichaelMCoates: So all applications, even ones created programmatically, have a manifestUrl. However, you're right, we could just delete the manifestUrl from the getInfo payload if the application was created programmatically and not pollute our LayoutApp interface. I'll make the change, thanks

src/demo/LayoutsUI.ts
```diff
@@ -174,25 +175,20 @@ export function createTabbedWindow(page: string) {
...
+            positionWindow(win);
+            // await ofWin.setBounds(win).catch((e: Error) => console.log('Setbounds error:', e));
         } else {
-            const ofWin = await fin.Window.wrap(win);
-            await ofWin.leaveGroup();
-            await ofWin.setBounds(win).catch((e: Error) => console.log('Setbounds error:', e));
+            // const ofWin = await fin.Window.wrap(win);
+            positionWindow(win);
+            // await ofWin.leaveGroup();
+            // await ofWin.setBounds(win).catch((e: Error) => console.log('Setbounds error:', e));
```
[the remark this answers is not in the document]
MichaelMCoates: Totally. I accidentally left the old implementation in case the new one failed. I'll change.

src/provider/workspaces/utils.ts
```diff
@@ -59,23 +55,27 @@ const createPlaceholder = async (win: WindowState) => {
...
+export const wasCreatedProgrammatically = (app: LayoutApp) => {
     return app && app.initialOptions && app.initialOptions.uuid && app.initialOptions.url;
 };
 
 // Type here should be ApplicationInfo from the js-adapter (needs to be updated)
 export const wasCreatedFromManifest = (app: LayoutApp, uuid?: string) => {
+    //@ts-ignore
     const {manifest, manifestUrl} = app;
     const appUuid = uuid || app.uuid;
     return typeof manifest === 'object' && manifest.startup_app && manifest.startup_app.uuid === appUuid;
```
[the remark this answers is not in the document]
MichaelMCoates: so this type here for app should be ApplicationInfo instead of LayoutApp from the js-adapter, which still needs to be updated. I could create a temporary interface for it above instead? 

As for this function in particular, it is called on the overall ApplicationInfo before save to determine whether or not an app was created from manifest. After this point, we use app.createdFromManifest in the rest of the code, but I'll change that per the first comment.

src/provider/workspaces/utils.ts
```diff
@@ -103,7 +103,5 @@ export const showingWindowInApp = async(app: LayoutApp): Promise<boolean> => {
 
 const isShowingWindow = async(ofWin: Window): Promise<boolean> => {
     const isShowing = await ofWin.isShowing();
-    const windowState = await ofWin.getState();
```
tgoc99: More of a design ? - do we want to include minimized windows in a layout?

tgoc99: more of a design question - do we want to include minimized windows in the layout?

src/demo/LayoutsUI.ts
```diff
@@ -174,25 +175,20 @@ export function createTabbedWindow(page: string) {
...
 async function onAppRes(layoutApp: LayoutApp): Promise<LayoutApp> {
-    await delayPromise(2000);
     console.log('Apprestore called:', layoutApp);
     const ofApp = await fin.Application.getCurrent();
     const openWindows = await ofApp.getChildWindows();
     const openAndPosition = layoutApp.childWindows.map(async (win, index) => {
         if(!openWindows.some((w: Application) => w.identity.name === win.name)) {
             const ofWin = await openChild(win.name, index, win.info.url);
-            await ofWin.setBounds(win).catch((e: Error) => console.log('Setbounds error:', e));
+            positionWindow(win);
```
tgoc99: need to await positionWindow?

src/demo/LayoutsUI.ts
```diff
@@ -174,25 +175,20 @@ export function createTabbedWindow(page: string) {
...
+            // await ofWin.setBounds(win).catch((e: Error) => console.log('Setbounds error:', e));
         } else {
-            const ofWin = await fin.Window.wrap(win);
-            await ofWin.leaveGroup();
-            await ofWin.setBounds(win).catch((e: Error) => console.log('Setbounds error:', e));
+            // const ofWin = await fin.Window.wrap(win);
+            positionWindow(win);
+            // await ofWin.leaveGroup();
```
tgoc99: need to await positionWindow?

src/provider/workspaces/utils.ts
MichaelMCoates: Hey! After a lot of discussion we have decided to include minimized windows in a layout (we didn't want a user thinking that one of their minimized windows that generates notifications is open when it's not). The future step is to give the user the power to filter out what items they want/don't want in the layout JSON that is being returned to them.

MichaelMCoates: See above

MichaelMCoates:
Addressed all comments and made code changes.

PR #63, opened 2018-08-21
[SERVICE-107] Integrate Snap&Dock with Save&Restore

tomrobertsOF:
- SnapService listens to native group events and updates the internal state to match
- minimal changes on the S&R side to deal with the fact that 'merge-group' doesn't raise an event when only two single windows are involved


src/provider/snapanddock/SnapService.ts
```diff
@@ -209,6 +211,49 @@ export class SnapService {
         }
     }
 
+    private onWindowGroupChanged(this: SnapService, event: fin.WindowGroupChangedEvent) {
```
pjbroadbent: Can we only add `this` to signatures if it's different to the default behaviour?

Since this method is in the `SnapService` class, TypeScript will assume `this` to be of type `SnapService` already.

src/provider/snapanddock/SnapWindow.ts
```diff
@@ -227,12 +233,16 @@ export class SnapWindow {
      * @param offset An offset to apply to this windows position (use this to enusre window is in correct position)
      * @param newHalfSize Can also simultaneously change the size of the window
      */
-    public setGroup(group: SnapGroup, offset?: Point, newHalfSize?: Point): void {
+    public setGroup(group: SnapGroup, offset?: Point, newHalfSize?: Point, synthetic?: boolean): void {
```
pjbroadbent: Add new parameter to tsdoc.

Would still like to remove the need for this in a future refactoring (if possible), but in the meantime needs to be properly documented.

src/provider/snapanddock/SnapWindow.ts
```diff
@@ -248,7 +258,7 @@ export class SnapWindow {
                 }
 
                 this.applyState(delta, () => this.snap());
```
pjbroadbent: Need a `synthetic` check on this `snap` call too?

test/demo/utils/executeJavascriptOnService.ts
```diff
@@ -8,7 +8,14 @@ import { Fin } from "hadouken-js-adapter";
 export async function executeJavascriptOnService(script: string): Promise<any> {
     const fin: Fin = await getConnection();
     return new Promise((resolve, reject) => {
-        fin.InterApplicationBus.subscribe({ uuid: "Layout-Manager", name: "Layout-Manager" }, "replytest", (message: string) => { console.log(message); resolve(message); })
+        fin.InterApplicationBus.subscribe({ uuid: "Layout-Manager", name: "Layout-Manager" }, "replytest", (message: string) => { 
+            console.log(message);
```
pjbroadbent: Remove the logging?

test/demo/utils/executeJavascriptOnService.ts
```diff
@@ -8,7 +8,14 @@ import { Fin } from "hadouken-js-adapter";
...
+            if (message.split(' ')[0] === 'Error:') {
```
pjbroadbent: Minor, but maybe define a message interface?

Rather than parsing the message, maybe pass `{success: boolean; message: string;}`?

test/provider/nativeGroupListeners.test.ts
```diff
@@ -0,0 +1,349 @@
...
+});
+test.afterEach.always(async () => {
+    if (win1 && win1.identity) { await win1.close(); }
+    if (win2 && win2.identity) { await win2.close(); }
+    win1 = win2 = {} as Window;
+});
+
+
+/* ====== 1 ====== */
+test('snap -> native ungroup win1 -> undock win1 >>> should not move ', async t => {
```
pjbroadbent: Test name definitely contains a lot of information, but I'm not sure what any of this means...

test/provider/nativeGroupListeners.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    await assertNotGrouped(win1, t);
+    await assertNotGrouped(win2, t);
+
+    // Undock
+    boundsBefore = await getBounds(win1);
+    await undockWindow(win1.identity as WindowIdentity);
+    boundsAfter = await getBounds(win1);
+
+    // Assert window did not move 
+    await assertNotMoved(boundsBefore, boundsAfter, t);
```
pjbroadbent: There's a lot of duplication in these tests, to the point where it's hard to see specifically what each one is actually testing.

Any way we can parameterise these tests, or extract some of the common parts?

test/provider/nativeGroupListeners.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    await win2.leaveGroup();
+    boundsAfter = await getBounds(win2);
+
+    // Assert window did not move 
+    await assertNotMoved(boundsBefore, boundsAfter, t);
+
+});
+
+test.failing('native group merge -> undock win1 -> native ungroup win1 >>> should move on undock ', async t => {
+// Runtime issue: native group merge does not raise an event when grouping two ungrouped windows
```
pjbroadbent: Can you raise this with Runtime and add the ticket number here?

test/provider/nativeGroupListeners.test.ts
```diff
@@ -0,0 +1,349 @@
...
+    // Window is alone in it's SnapGroup
+    const snapGroup = await getSnapGroup(win);
+    t.is(snapGroup.length, 1);
+}
+
+async function assertMoved(bounds1:NormalizedBounds, bounds2:NormalizedBounds, t: TestContext) {
+    t.notDeepEqual(bounds1, bounds2);
+}
+
+async function assertNotMoved(bounds1:NormalizedBounds, bounds2:NormalizedBounds, t: TestContext) {
```
pjbroadbent: The above utils are synchronous, any reason they're all tagged as `async`?

res/test/provider.html
```diff
@@ -8,22 +8,45 @@
...
-            } else {
-                fin.desktop.InterApplicationBus.send("TEST", "replytest", "script evaluated");
+            fin.desktop.System.showDeveloperTools()
+
+            try {
+                const returnVal = eval(message);
+
+                if (returnVal instanceof Promise) {
+                    returnVal.then((val) => {
+                        if (val) {
```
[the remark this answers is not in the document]
tomrobertsOF: Good point. I'll fix that.

src/provider/snapanddock/SnapService.ts
tomrobertsOF: Oh yea, sorry. Removing it now

res/test/provider.html
```diff
@@ -8,22 +8,45 @@
...
+                } else if (returnVal) {
+                    try {
+                        JSON.stringify(val);
+                        fin.desktop.InterApplicationBus.send("TEST", "replytest", returnVal);
+                    } catch (error) {
+                        fin.desktop.InterApplicationBus.send("TEST", "replytest", 'Error: ' + error);
+                    }
+                } else {
+                    fin.desktop.InterApplicationBus.send("TEST", "replytest", "Script executed");
+                }
```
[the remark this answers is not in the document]
tomrobertsOF: Done

test/provider/nativeGroupListeners.test.ts
tomrobertsOF: Issue may not have been the right choice of words. This is runtime behaviour.

test/provider/nativeGroupListeners.test.ts
tomrobertsOF: When writing the functions originally I wasn't sure if I would need to await any async calls, so put them as async. Turns out I don't so I'll remove.

tomrobertsOF:
executeJavascriptOnService has some major issues in this iteration. I will close this PR, roll some stuff back, address the review comments and make a new PR.

src/provider/snapanddock/SnapWindow.ts
tomrobertsOF: Didn't see that one. It wasn't noticeable in my testing because the current implementation only uses synthetic with undefined for offset and newHalfSize. That may not always be the case, so I've added a guard like the others

test/provider/nativeGroupListeners.test.ts
tomrobertsOF: Possibly, they are basically a load of permutations of the same thing, but slightly different. I'll have a think about it.

PR #64, opened 2018-08-22
[SERVICE-107] Integrate Snap&Dock with Save&Restore

tomrobertsOF:
- SnapService listens to native group events and updates the internal state to match (deregistered windows are ignored).
- Minimal changes on the S&R side to deal with the fact that 'merge-group' doesn't raise an event when only two single windows are involved

Automated testing is in place for SnapGroup <-> Native group correspondence. 
Manual testing will be needed for restoring layouts for the time being.


tomrobertsOF:
Manual testing completed and no issues found. Waiting for review to merge.

PR #65, opened 2018-08-22
SERVICE-212 - fix npm publishing, project consistency

bschwinn:
making consistent with fdc/notifs.  tsc and webpack are now separate, the demo apps now use the client from the ./dist directory via package.json.  Unfortunately I had to create another package.json and webpack.config.js to keep things separate (see commits for chicken/egg scenario).


pbaize:
I'm not convinced npm installing the local client is the best solution here though. Seems like an unnecessary step

src/demo/LayoutsUI.ts
```diff
@@ -205,13 +206,24 @@ function removeForgetWins(window: ServiceIdentity) {
 }
 
 function addLayoutNamesToDropdown() {
+<<<<<<< HEAD
```
MichaelMCoates: Merge conflict! We want the Storage.getAlllayoutIDs code.

MichaelMCoates:
Approving the updated code.

PR #78, opened 2018-08-24
Fix demo windows not persisting frames on restore

tomrobertsOF:
Issue reported where saving and restoring the frameless snappable windows from the demo does not maintain the framelessness of the windows.


tomrobertsOF:
Agreed. This is a purely temporary measure to improve UX in the demo. Will be changed in a demo/manager refactor down the line

PR #79, opened 2018-08-24
[ SERVICE-224 ] Move Storage Layer in S&R from the Provider to the Manager

MichaelMCoates:


MichaelMCoates:
Good catch, updated.

src/demo/LayoutsUI.ts
```diff
@@ -19,7 +20,8 @@ window.forgetMe = forgetMe;
 export async function setLayout() {
     const name = (document.getElementById('layoutName') as HTMLTextAreaElement).value;
     const layoutSelect = document.getElementById('layoutSelect') as HTMLSelectElement;
-    const layout = await Layouts.saveCurrentLayout(name);
+    const layout = await Layouts.generateLayout();
+    layout.name = name;
```
pjbroadbent: Should we remove `name` from the `Layout` object? We no longer use it anywhere in the service, and clients will likely want to manage how they identify layouts in their own way.

MichaelMCoates: Yep, let's do it.

MichaelMCoates:
Removed `name` from the Layout interface. Moved to a Workspace interface

tomrobertsOF:
My review still stands. Happy to merge it.

PR #80, opened 2018-08-25
[SERVICE-221] Tab APIs

ryanpetersonOF:
Updated Tabbing APIs to the API v0.9 spec


src/client/main.ts
```diff
@@ -215,8 +215,8 @@ export async function getTabs(window: Identity = getId()): Promise<Identity[]|nu
  * If a custom tab-strip UI is being used - this sets the URL for the tab-strip.
  * This binding happens on the application level.  An application cannot have different windows using different tabbing UI.
  */
-export async function setTabClient(url: string, config: TabClientConfig): Promise<void> {
-    if (!config || isNaN(config.height)) {
+export async function setTabClient(url: string, config: TabWindowOptions): Promise<void> {
```
pjbroadbent: Is `TabWindowOptions` the right interface? The client is expected to pass a `TabWIndowOptions` (which contains members that don't seem relevant here), but the provider receives an `ApplicationUIConfig`?..

PR #83, opened 2018-08-28
[SERVICE-228] Certain windows will occasionally fail to register with the S&D service

tomrobertsOF:
Fix a bug where new windows would occasionally not be registered with the service.

### DIFF
--- a/src/provider/snapanddock/SnapService.ts
+++ b/src/provider/snapanddock/SnapService.ts
@@ -70,16 +70,26 @@ export class SnapService {
 
         const serviceId: string = fin.desktop.Application.getCurrent().uuid;
 
-        fin.desktop.System.addEventListener('application-created', (event: fin.SystemBaseEvent) => {
-            console.log('New application created: ' + event.uuid);
-            this.registerApplication(event.uuid);
+        // Listen for any new windows created and register them with the service
+        fin.desktop.System.addEventListener('window-created', (event: fin.WindowBaseEvent) => {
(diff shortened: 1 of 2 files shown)


tomrobertsOF:
For some reason, there are whitespace changes in `fin.d.ts`. I ran gts fix and everything came back fine. Not sure how they didn't come up before.

src/provider/snapanddock/SnapService.ts
```diff
@@ -156,23 +166,16 @@ export class SnapService {
...
-        // Listen for future windows
-        app.addEventListener('window-created', (event: fin.WindowEvent) => {
-            this.addWindow(fin.desktop.Window.wrap(event.uuid, event.name)).then(window => console.log('App created new window: ' + window.getId()));
-        });
+        // Check that the service does not already have a matching window
+        const existingSnapWindow = this.getSnapWindow(newOFWindow);
+        if (!existingSnapWindow) {
+            this.addWindow(newOFWindow).then((win: SnapWindow) => console.log('Registered window: ' + win.getId()));
+        } else {
+            console.warn('Attempted to register an already existing window: ', existingSnapWindow.getId());
```
pjbroadbent: In this scenario, `existingSnapWindow` likely has stale/invalid state.

Would it make more sense to destroy `existingSnapWindow`, and then continue with `this.addWindow` as if there wasn't an existing window?

tomrobertsOF: Done

PR #84, opened 2018-08-28
[SERVICE-229] Change onWillSaveAppLayout to only pass customData back

MichaelMCoates:
We want users to only be able to add customData to a Layout, not change the layout information on the fly.

### DIFF
--- a/src/client/main.ts
+++ b/src/client/main.ts
@@ -115,7 +115,7 @@ export async function addEventListener(
  * Decide which parts of this you will implement, alter LayoutApp object to reflect this then send it back
  */
 // tslint:disable-next-line:no-any
-export async function onWillSaveAppLayout(customDataDecorator: () => any): Promise<boolean> {
+export async function onApplicationSave(customDataDecorator: () => any): Promise<boolean> {
     const service: ServiceClient = await servicePromise;
     return service.register('savingLayout', customDataDecorator);
 }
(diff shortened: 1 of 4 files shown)


src/client/main.ts
```diff
@@ -114,9 +114,10 @@ export async function addEventListener(
 /**
  * Decide which parts of this you will implement, alter LayoutApp object to reflect this then send it back
  */
-export async function onWillSaveAppLayout(layoutDecorator: (layoutApp: LayoutApp) => LayoutApp | false | Promise<LayoutApp|false>): Promise<boolean> {
+// tslint:disable-next-line:no-any
+export async function onApplicationSave(customDataDecorator: () => any): Promise<boolean> {
```
pjbroadbent: Can we add a type alias for the custom data? Something like `export type CustomData = any;`. Won't give any extra type safety, but will provide some extra semantic info. Also means that we'll only need one `no-any` comment (on the `CustomData` declaration), rather than every time we need to refer to `customData`.

Can the comment also clarify the return value? What's the difference between `resolve(false)` and `reject()`.

MichaelMCoates: Great idea, updated to include the CustomData type. 

The return value is the result of the register() call. I think it can either reject or resolve(true): https://github.com/HadoukenIO/js-adapter/blob/develop/src/api/interappbus/channel/channel.ts#L107

Should we be documenting that here?

PR #89, opened 2018-08-29
[SERVICE-213] Deregistered windows still have opacity effect applied when being dragged

tomrobertsOF:
Fix window opacity changing when dragging deregistered windows


espenove:
Manual testing looks good

PR #96, opened 2018-08-30
[SERVICE-226] - Switch tab strip to use new API

JohnathonLamOF:


src/provider/tabbing/tabstrip/TabManager.ts
```diff
@@ -54,7 +60,8 @@ export class TabManager {
                     return {uuid: el.dataset.uuid as string, name: el.dataset.name as string};
                 });
                 // Sends the new order to the service to update the cache
-                TabManager.tabAPI.sendTabOrder(orderedTabList);
+                //TabManager.tabAPI.sendTabOrder(orderedTabList);
```
tomrobertsOF: commented out code

PR #99, opened 2018-08-30
Blank Window Fix

ryanpetersonOF:
* Adds Xaviers work around for blank windows
* Removes wait on tab strip showing


tomrobertsOF:
Will this have any noticeable side effects from a user's perspective?

ryanpetersonOF:
It will show the tab strip quicker with a brief white flash, better than the current frameless window then tab window.  Will be changed later today overall with the new windowing strategy.

Xaviers fix I don't believe will have any noticeable effects.  If they do it will be less than the current blank window experience...

PR #103, opened 2018-08-31
[SERVICE-239] Maximized Windows + Tabbing Improvements

ryanpetersonOF:
Fixes some maximize + tabbing logic

### DIFF
--- a/src/provider/tabbing/GroupWindow.ts
+++ b/src/provider/tabbing/GroupWindow.ts
@@ -123,7 +123,7 @@ export class GroupWindow extends AsyncWindow {
      */
     public async restoreGroup(): Promise<void|void[]> {
         if (this._isMaximized) {
-            if (await this._tabGroup.activeTab.window.getState()) {
+            if (await this._tabGroup.activeTab.window.getState() === 'maximized') {
                 this._isMaximized = false;
                 Promise.all(this._tabGroup.tabs.map(tab => tab.window.restore()));
                 return this._tabGroup.hideAllTabsMinusActiveTab();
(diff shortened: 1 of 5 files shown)


src/provider/tabbing/GroupWindow.ts
```diff
@@ -123,8 +123,10 @@ export class GroupWindow extends AsyncWindow {
      */
     public async restoreGroup(): Promise<void|void[]> {
         if (this._isMaximized) {
-            if ((await this.getState()) === 'minimized') {
-                return this._tabGroup.activeTab.window.restore();
+            if (await this._tabGroup.activeTab.window.getState()) {
```
tomrobertsOF: Should this not still have the `=== minimised`, or is there some other logic hidden away that handles that?

src/provider/tabbing/TabService.ts
```diff
@@ -164,18 +164,23 @@ export class TabService {
 
         const firstTab = tabsP.shift();
 
-        if (firstTab) {
-            const bounds = await firstTab.window.getWindowBounds();
-            tabsP.forEach(tab => tab.window.finWindow.setBounds(bounds.left, bounds.top, bounds.width, bounds.height));
-            tabsP[tabsP.length - 1].window.finWindow.bringToFront();
-            await group.addTab(firstTab, false);
+        if (!firstTab) {
+            return Promise.reject('Something is wrong...');
```
tomrobertsOF: Can you make this a more useful error message.

ryanpetersonOF: It is actually impossible to reach this.  We check if the if array has length right away.  Shift() returns undefined on a 0 length array

tomrobertsOF: If it will never get called, why is it there?

PR #104, opened 2018-08-31
Slight UX Tweaks to S&D

tomrobertsOF:
Slight changes snap and dock UX to address usability concerns.

### DIFF
--- a/src/provider/snapanddock/SnapPreview.ts
+++ b/src/provider/snapanddock/SnapPreview.ts
@@ -114,7 +114,7 @@ export class SnapPreview {
             frame: false,
             state: 'normal',
             autoShow: false,
-            alwaysOnTop: true
+            alwaysOnTop: false
         };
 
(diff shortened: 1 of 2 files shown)


src/provider/snapanddock/SnapPreview.ts
```diff
@@ -114,7 +114,7 @@ export class SnapPreview {
             frame: false,
             state: 'normal',
             autoShow: false,
-            alwaysOnTop: true
+            alwaysOnTop: false
```
pjbroadbent: May need a bring to front adding somewhere? This window persists, so could easily end up behind other windows, especially non-of windows

espenove: We are aware of the windows sometimes getting stuck behind non-openfin windows, the scenarios for the window being behind is fairly small, and the effect better.
Might need to be looked at later given feedback

PR #106, opened 2018-08-31
[SERVICE-241] Bring all other windows in group to front when snapped window is focused

tomrobertsOF:
Added onFocused listener to each SnapWindow which will bring all other windows in it's group to front

### DIFF
--- a/src/provider/snapanddock/SnapWindow.ts
+++ b/src/provider/snapanddock/SnapWindow.ts
@@ -425,24 +425,23 @@ export class SnapWindow {
         }
     };
     private handleFocused = async () => {
-        // Loop through all windows in the same group as the focused window and set them all to 
+        // Loop through all windows in the same group as the focused window and set them all to
         // alwaysOnTop: true, then set them all to alwaysOnTop: false. This will bring them to front
         // without triggerig further events, thereby avoiding an event loop.
-           this.window.getGroup(async (group: fin.OpenFinWindow[]) => {
-               const numWindows = group.length;
-               // Set always on top (brings windows to front without triggering another event)
-               for (let i = 0; i < numWindows; i++) {
-                   await new Promise((res, rej) => {
-                      group[i].updateOptions({alwaysOnTop: true}, res, rej);
-                   });
-               }
-               // Remove always on top (windows stays on top) 
-               for (let i = 0; i < numWindows; i++) {
-                   await new Promise((res, rej) => {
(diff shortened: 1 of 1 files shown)


tomrobertsOF:
I wouldn't think it should have any effect at all, but if it does it wouldn't be too much to add some special handling once we have a better idea of what the interactions will look like.

src/provider/snapanddock/SnapWindow.ts
```diff
@@ -422,4 +424,24 @@ export class SnapWindow {
...
+            // Set always on top (brings windows to front without triggering another event)
+            for (let i = 0; i < numWindows; i++) {
+                await new Promise((res, rej) => {
+                    group[i].updateOptions({alwaysOnTop: true}, res, rej);
+                });
+            }
+            // Remove always on top (windows stays on top)
+            for (let i = 0; i < numWindows; i++) {
+                await new Promise((res, rej) => {
+                    group[i].updateOptions({alwaysOnTop: false}, res, rej);
```
pjbroadbent: Could any of these windows have had alwaysOnTop set previously?

What event is it that we're attempting to avoid? Does bringing a window to front trigger a focused event?

tomrobertsOF: I did it with bringToFront at Xavier's recommendation, but having just tried it with bringToFront it seems to work ok. I can push it up with that method for now, and check with him later on to see if there was any specific reason for doing it that way?

src/provider/snapanddock/SnapWindow.ts
```diff
@@ -422,4 +424,18 @@ export class SnapWindow {
...
+        this.window.getGroup(async (group: fin.OpenFinWindow[]) => {
+            const promises: Promise<void>[] = [];
+            for (let i = 0; i < group.length; i++) {
+                promises[i] = new Promise<void>((res, rej) => {
+                    group[i].bringToFront(res, rej);
+                });
+            }
+
+            await Promise.all(promises);
+        });
```
ryanpetersonOF: This could be shortened with a `.map`:

```
        this.window.getGroup(async (group: fin.OpenFinWindow[]) => {
            await Promise.all(group.map(win => new Promise((res, rej) => {win.bringToFront(res,rej)})));
        });
```

tomrobertsOF: It could, but a vanilla for-loop is generally a fair bit more performant than map, and since we are messing with window visibility, I'd rather leave as is.

PR #107, opened 2018-08-31
[SERVICE-184] Integrate Tab Set restoration into Save and Restore

MichaelMCoates:


src/provider/workspaces/restore.ts
```diff
@@ -6,7 +6,13 @@ import {Layout, LayoutApp, LayoutName, WindowState} from '../../client/types';
 
 import {regroupLayout} from './group';
 import {providerChannel} from '../main';
-import {createAppPlaceholders, isClientConnection, positionWindow, wasCreatedProgrammatically} from './utils';
+import {createAppPlaceholders, createTabPlaceholder, createNormalPlaceholder, isClientConnection, positionWindow, wasCreatedProgrammatically} from './utils';
+
+import {createTabGroupsFromTabBlob} from '../tabbing/TabUtilities';
+import { TabService } from '../tabbing/TabService';
+import { removeTab } from '../tabbing/SaveAndRestoreAPI';
```
pjbroadbent: Run the linter

src/client/types.ts
```diff
@@ -42,6 +42,7 @@ export interface Layout {
     monitorInfo: any;  // saving but not using yet
     customData?: any;
     apps: LayoutApp[];
+    tabGroups?: TabBlob[];
```
pjbroadbent: Is this really optional? Seems it should always be there, array will just be empty if there's no tab sets

src/provider/tabbing/SaveAndRestoreAPI.ts
```diff
@@ -71,6 +71,27 @@ export async function swapTab(add: TabIdentifier, swapWith: TabIdentifier) {
     return;
 }
 
+/**
+ * Removes a tab from a tab group.
+ * @param {TabIdentifier} tabID The identity of the tab to remove.
+ */
+export async function removeTab(tabID: TabIdentifier) {
```
pjbroadbent: Is this needed? Seems anything calling this could just call the 'actual' removeTab?

src/provider/workspaces/restore.ts
```diff
@@ -43,7 +49,127 @@ export const restoreApplication = async(layoutApp: LayoutApp, resolve: Function)
 export const restoreLayout = async(payload: Layout, identity: Identity): Promise<Layout> => {
     const layout = payload;
     const startupApps: Promise<LayoutApp>[] = [];
-    console.log('Restoring layout:', layout);
+    /*tslint:disable-next-line:no-any*/
+    const tabbedWindows: any = {};
+    /*tslint:disable-next-line:no-any*/
+    const tabbedPlaceholdersToWindows: any = {};
```
pjbroadbent: You can type these rather than using any.

```
cost tabbed...: {[uuid: string]: {[name: string]: WindowIdentity}};
```
Think that's right, on mobile. There's also a built-in Map type. Not a drop-in replacement like the above declaration, though.

src/provider/workspaces/restore.ts
MichaelMCoates: Ran it, fixed below.

src/client/types.ts
MichaelMCoates: Totally agree, I'll add it in

src/provider/tabbing/SaveAndRestoreAPI.ts
ryanpetersonOF: `group.removeTab` requires you to know the group first.  We can make a more top-level removeTab later on.

PR #112, opened 2018-09-03
[SERVICE-242] S&R now saves and restores window framedness

tomrobertsOF:
- Updated save/restore data model to include a boolean denoting whether window has a frame
 - Removed workaround code from the demo

NOTE: Restoring a layout saved before this change, will result in all windows being restored with a frame. Any layouts saved from this point forward will preserve the state as expected.


pjbroadbent:
@MichaelMCoates Do we not need to know this within the service for creating the child window previews?

Also, whilst it may not be strictly necessary, I'm not sure there's any harm in providing extra information to applications during restore. The "workaround" that this PR also removes from `src/demo/LayoutsUI.ts` could be a good example of where the extra info could be useful.

MichaelMCoates:
RE: Window previews: That's a good point. At the moment, we don't change the placeholder window frame based on the option. Actually, it might make the most sense to make all placeholder windows frameless so that a user can't drag them around. Not sure what the best course of action is.

Totally agree, I'm just trying to follow in the trend we've had so far of keeping the schema as limited as possible. My thought is that most users won't be calling a normal `new fin.desktop.Window({})` to spin up child windows, but instead calling one of their own internal functions to launch a pop-up/child window programmatically, with their own specifications. But yeah it would be nice to have that information readily available. I defer to @nkolba on this one.

Lastly, if we did include this data, I'm not sure `positionWindow` is the best place to put the logic for adding/disabling the frame.

espenove:
Hi @MichaelMCoates  - I agree with @pjbroadbent on this, no harm having this info, and will let us create proper preview windows

PR #132, opened 2018-09-04
Fix windowToGroup joinGroup behavior. Ignore layouts-service uuids and catch errors correctly.

MichaelMCoates:


pjbroadbent:
Ryan is AFK, but approves changes too.

PR #141, opened 2018-09-05
replace git sha with version in cdn folder structure

bschwinn:


Jenkinsfile
```diff
@@ -50,20 +50,22 @@ pipeline {
                 script {
                     GIT_SHORT_SHA = sh ( script: "git rev-parse --short HEAD", returnStdout: true ).trim()
                     VERSION = sh ( script: "node -pe \"require('./package.json').version\"", returnStdout: true ).trim()
-                    S3_LOC = env.DSERVICE_S3_ROOT + "layouts/" + GIT_SHORT_SHA
-                    PROD_JSON = env.DSERVICE_S3_ROOT + "layouts/" + "app-" + VERSION + ".json"
+                    S3_LOC = env.DSERVICE_S3_ROOT + "layouts/" + VERSION
+                    PROD_JSON = env.DSERVICE_S3_ROOT + "layouts/app.json"
+                    PROD_VERSIONED_JSON = env.DSERVICE_S3_ROOT + "layouts/app-" + VERSION + ".json"
```
pjbroadbent: Do we still need this if the directories are named after the versions?

Is it worth having both `.../layouts/app-0.9.0.json` and `.../layouts/0.9.0/app.json`?

bschwinn: not really sure, now with the directories being much more friendly it may be redundant.

pjbroadbent: I vote to drop. Definitely redundant, IMO.

bschwinn: I don't think this should hold up this change, if we want to omit the creation of this file, it can be done in a new nice n small PR.

bschwinn: i'll remove in the spirit of less is more

PR #142, opened 2018-09-10
[SERVICE-191] Update to v34 and migrate service to IAB Channels 

tomrobertsOF:
Updated all apps to use runtime 9.61.34.22
Updated hadouken-js-adapter to version 0.34.4
Moved all provider code from services API to IAB.channels api
Improved type safety throughout

### DIFF
--- a/src/provider/main.ts
+++ b/src/provider/main.ts
@@ -1,4 +1,3 @@
-import {Provider} from 'hadouken-js-adapter/out/types/src/api/services/provider';
 import {Identity} from 'hadouken-js-adapter/out/types/src/identity';
 
 import {TabAPI} from '../client/APITypes';
@@ -10,12 +9,14 @@ import {win10Check} from './snapanddock/utils/platform';
 import {TabService} from './tabbing/TabService';
 import {generateLayout} from './workspaces/create';
 import {getAppToRestore, restoreApplication, restoreLayout} from './workspaces/restore';
+import { ChannelProvider } from 'hadouken-js-adapter/out/types/src/api/interappbus/channel/provider';
+import { ProviderIdentity } from 'hadouken-js-adapter/out/types/src/api/interappbus/channel/channel';
 
 export let snapService: SnapService;
 export let tabService: TabService;
-export let providerChannel: Provider;
+export let providerChannel: ChannelProvider;
 declare const window: Window&{
-    providerChannel: Provider;
+    providerChannel: ChannelProvider;
     snapService: SnapService;
     tabService: TabService;
 };
@@ -23,7 +24,7 @@ declare const window: Window&{
 fin.desktop.main(main);
 
 async function registerService() {
-    providerChannel = window.providerChannel = (await fin.desktop.Service.register()) as Provider;
+    providerChannel = window.providerChannel = await fin.InterApplicationBus.Channel.create();
     providerChannel.onConnection((app, payload) => {
         if (payload && payload.version && payload.version.length > 0) {
             console.log(`connection from client: ${app.name}, version: ${payload.version}`);
@@ -97,9 +98,13 @@ export async function main() {
  * @param {SnapWindow} window The target to which the message will be sent
  * @param {fin.OpenFinServiceProvider} provider Provider object wrapping an instance of the openfin layouts service
  */
-function sendWindowServiceMessage(action: GroupEventType, window: SnapWindow, provider: fin.OpenFinServiceProvider) {
-    console.log('Dispatching window message: ', action, 'to window: ', window.getIdentity());
-    provider.dispatch(window.getIdentity(), action, {});
+function sendWindowServiceMessage(action: GroupEventType, window: SnapWindow, provider: ChannelProvider) {
+    const {uuid, name} = window.getIdentity();
+    const to: ProviderIdentity | undefined = provider.connections.find(conn => conn.uuid === uuid && conn.name === name);
+    if (to) {
+        console.log('Dispatching window message: ', action, 'to window: ', window.getIdentity());
+        provider.dispatch(to, action, {});
+    }
 }
 
 /**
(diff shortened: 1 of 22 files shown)


src/provider/main.ts
```diff
@@ -10,20 +9,22 @@ import {win10Check} from './snapanddock/utils/platform';
 import {TabService} from './tabbing/TabService';
 import {generateLayout} from './workspaces/create';
 import {getAppToRestore, restoreApplication, restoreLayout} from './workspaces/restore';
+import { ChannelProvider } from 'hadouken-js-adapter/out/types/src/api/interappbus/channel/provider';
+import { ProviderIdentity } from 'hadouken-js-adapter/out/types/src/api/interappbus/channel/channel';
```
pjbroadbent: Needs a lint?..

src/provider/snapanddock/SnapService.ts
```diff
@@ -102,8 +102,7 @@ export class SnapService {
             .register(
                 'CommandOrControl+Shift+U',
                 () => {
-                    // @ts-ignore - v2api types missing
-                    fin.System.getFocusedWindow().then(focusedWindow => {
```
pjbroadbent: Are the types still missing in the v34 hadouken-js lib?

src/provider/workspaces/create.ts
```diff
@@ -92,7 +92,13 @@ export const generateLayout = async(payload: null, identity: Identity): Promise<
             console.log('Connected application', app.uuid);
 
             // HOW TO DEAL WITH HUNG REQUEST HERE? RESHAPE IF GET NOTHING BACK?
-            let customData = await providerChannel.dispatch({uuid: app.uuid, name: app.uuid}, 'savingLayout', app);
+            // tslint:disable-next-line:no-any
+            let customData: any = undefined;
```
pjbroadbent: There's a `CustomData` type to use for this purpose. It's still `any`, but avoids having to add the tslint comments in every place we refer to custom data.

src/provider/tabbing/Tab.ts
```diff
@@ -62,16 +64,17 @@ export class Tab {
...
+        const payload = {tabGroupID: this.tabGroup.ID, tabID: this.ID, tabProps: this._tabProperties, index: this.tabGroup.getTabIndex(this._tabID)};
+
+        const tabConnection = getClientConnection(this.ID);
+        if (tabConnection) {
+            this.mService.dispatch(tabConnection, 'join-tab-group', payload);
+        }
+        const tabStripConnection = getClientConnection({uuid: TabServiceID.UUID, name: this.tabGroup.ID});
+        if (tabStripConnection) {
+            this.mService.dispatch(tabStripConnection, 'join-tab-group', payload);
+        }
```
pjbroadbent: A lot of boilerplate being added, each one-line `dispatch` call is now four lines.

Can we re-work the util function? Make this something more like `sendToClient(this.ID, 'join-tab'group', payload)` - with the null-checking happening within the util function?

tsconfig.json
```diff
@@ -9,9 +9,11 @@
     },
     "include": [
         "src/fin.d.ts",
+        "src/finV2.d.ts",
         "src/provider/snapanddock/**/*.ts",
         "src/provider/tabbing/**/*.ts",
         "src/provider/workspaces/**/*.ts",
-        "src/client/main.ts"
+        "src/client/main.ts",
+        "src/demo/LayoutsUI.ts"
```
pjbroadbent: Is this an intentional change? How was the demo code being built previously? And should be the whole demo dir, rather than just the one file?

espenove:
Works like a charm

src/provider/snapanddock/SnapService.ts
tomrobertsOF: They are currently a bit questionable. Typing says `getFocusedWindow` returns a `Promise<WindowInfo>`, but it seems like this may be a mistake as when you call it from the console it actually returns `Promise<Identity | null>`.

src/provider/workspaces/create.ts
tomrobertsOF: Ah, ok. It was any typed when i got there, so must have been missed when adding in the type. Fixed in next commit.

src/provider/tabbing/Tab.ts
tomrobertsOF: The util function is used in the S&R code as is, but i can make an extra util to wrap this logic?

tsconfig.json
tomrobertsOF: Demo code was being built, but this is required to inject the types correct types into it. Previously it just used `declare var fin:any;` at the top of the file and had no real type safety on fin operations.

Now that we have v1 and v2 types working together, I thought it made sense to add type checking to this file.

tomrobertsOF:
@MichaelMCoates do you have any steps to repro reliably? Not been able to break it on my side.

src/provider/tabbing/Tab.ts
tomrobertsOF: Also, should probably add that `ChannelProvider.dispatch` has been changed in v35 to just take an identity object, so we can do away with the util once we get there.

src/provider/workspaces/utils.ts
```diff
@@ -144,9 +146,10 @@ interface AppInfo {
 }
 
 // Type here should be ApplicationInfo from the js-adapter (needs to be updated)
-export const wasCreatedFromManifest = (app: AppInfo, uuid?: string) => {
-    const {manifest, manifestUrl} = app;
-    const appUuid = uuid || app.uuid;
+export const wasCreatedFromManifest = (app: ApplicationInfo, uuid?: string) => {
+
+    const {manifest} = {...app, uuid} as AppInfo;
+    const appUuid = uuid || '';
```
MichaelMCoates: I think I'm missing something, but why did you change this to empty string?

tomrobertsOF:
@MichaelMCoates, @pjbroadbent, the register/deregister race condition should now be resolved. Can you retest and confirm?

PR #143, opened 2018-09-11
[SERVICE-250] Can tab to minimised windows

ryanpetersonOF:
Fixes the 'feature' to tab to a minimized window.

### DIFF
--- a/src/provider/snapanddock/SnapService.ts
+++ b/src/provider/snapanddock/SnapService.ts
@@ -385,7 +385,7 @@ export class SnapService {
                 const windowUnderPoint = getWindowAt(activeState.center.x, activeState.center.y, currentDragWindowIdentity);
 
                 // There is a window under our drop point
-                if (windowUnderPoint) {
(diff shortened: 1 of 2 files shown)


src/provider/snapanddock/SnapService.ts
```diff
@@ -385,7 +385,7 @@ export class SnapService {
                 const windowUnderPoint = getWindowAt(activeState.center.x, activeState.center.y, currentDragWindowIdentity);
 
                 // There is a window under our drop point
-                if (windowUnderPoint) {
+                if (windowUnderPoint && this.getSnapWindow(windowUnderPoint)!.getState().state !== 'minimized') {
```
tomrobertsOF: I think this would make more sense inside `getWindowAt`. Can we really say there is a window at a location if that window has since been minimised? Also, having the check inside would cut out any chance of this happening in any other case where getWindowAt is called.

ryanpetersonOF: Good Idea.  I have updated it now

tomrobertsOF: Looks good!

PR #144, opened 2018-09-12
[SERVICE-251] When maximising snapped windows, sometimes other windows are brought in front of maximised window on focus

tomrobertsOF:
Bug fix. Windows will not stay behind the maximised window when it is focused


tomrobertsOF:
Thought of a couple of improvements that could be made. Will close this and reopen once I've finished those.

PR #146, opened 2018-09-14
[SERVICE-251] When maximising snapped windows, sometimes other windows are brought in front of maximised window on focus

tomrobertsOF:
Fixed issue with z-indexing on maximised windows in snap groups. Other windows in the group will now remain behind the maximised window unless they themselves are focused.

Also cleaned up the eventing code in SnapWindow and generally improved they type definitions for the openfin event listener functions while I was in there.

### DIFF
--- a/src/provider/snapanddock/SnapWindow.ts
+++ b/src/provider/snapanddock/SnapWindow.ts
@@ -40,6 +40,8 @@ export enum eTransformType {
     RESIZE = 1 << 1
 }
 
+type OpenFinWindowEventHandler = <K extends keyof fin.OpenFinWindowEventMap>(event: fin.OpenFinWindowEventMap[K]) => void;
+
 export class SnapWindow {
     public static async getWindowState(window: fin.OpenFinWindow): Promise<WindowState> {
         return Promise
@@ -119,7 +121,7 @@ export class SnapWindow {
 
     // Tracks event listeners registered on the fin window for easier cleanup.
     private registeredListeners:
-        Map<keyof fin.OpenFinWindowEventMap, (<K extends keyof fin.OpenFinWindowEventMap>(event: fin.OpenFinWindowEventMap[K]) => void)[]> = new Map();
+        Map<keyof fin.OpenFinWindowEventMap, OpenFinWindowEventHandler[]> = new Map();
 
     // State tracking for "synth move" detection
     private boundsChangeCountSinceLastCommit: number;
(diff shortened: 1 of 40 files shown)


src/provider/snapanddock/SnapWindow.ts
```diff
@@ -344,20 +350,27 @@ export class SnapWindow {
         }
     }
 
+    private registerListener<K extends keyof fin.OpenFinWindowEventMap>(eventType: K, handler: (event: fin.OpenFinWindowEventMap[K]) => void) {
+        this.window.addEventListener(eventType, handler);
```
pjbroadbent: Can we turn the handlers back into functions now? Rather than using the "member variable and arrow syntax" method?

Either as regular private functions with a `.bind` here, or by putting them back as inline arrow functions?..

src/provider/snapanddock/SnapWindow.ts
```diff
@@ -344,20 +350,27 @@ export class SnapWindow {
...
-            this.window.removeEventListener('closed', this.handleClosed);
-            this.window.removeEventListener('bounds-changing', this.handleBoundsChanging);
-            this.window.removeEventListener('focused', this.handleFocused);
+            console.log('OnClose recieved for window ', this.getId(), '. Removing listeners');
+
+            for (const [key, listenerArray] of new Map(this.registeredListeners)) {
+                for (const listener of listenerArray) {
+                    this.window.removeEventListener(key, listener);
+                }
+                this.registeredListeners.delete(key);
```
pjbroadbent: Rather than cloning the map and deleting items one-by-one, would it make more sense to iterate over the map and then `clear()` it after the loop?

src/provider/snapanddock/SnapWindow.ts
```diff
@@ -425,17 +438,26 @@ export class SnapWindow {
         }
     };
     private handleFocused = async () => {
```
pjbroadbent: I don't think the promises in this function work as intended?.. The `getGroup` callback awaits on the `bringToFront` promises, but there's nothing to propagate that up to the level of `handleFocused`.

Also, since `handleFocused` is only ever used as an event listener, having it return a promise would have no effect anyway?..

tomrobertsOF:
The `this.state.state` part is because the SnapWindow object has a `state` property which encapsulates all of the relevant information about the OpenFin Window. This includes the OpenFin window `state` (minimized/maximized/...). It's a bit weird i'll admit, but makes sense in the context.

PR #148, opened 2018-09-17
Parameterize test runner

tomrobertsOF:
Allow passing test file and name filters through npm run as command line arguments.

  `--file-path [String]` : Specifies the name of the file containing the tests to run. 
      Example: `--file-path undock` will run tests in the file `undock.test.ts`

  `--filter [String]` : Filters the tests that will be run. 
      Valid filter syntax is described in the [ava documentation](https://github.com/avajs/ava#running-tests-with-matching-titles).
      Example: --filter *vertical* will run all tests containing the word 'vertical'

  Any other command line parameters will be passed through to ava as-is. A list of valid command line parameters can be found in the [ava documentation](https://github.com/avajs/ava#cli). (NOTE: `--match` is not supported, use `--filter` instead).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6316,12 +6316,13 @@
       }
     },
     "hadouken-js-adapter": {
-      "version": "0.32.1-alpha.5",
-      "resolved": "https://registry.npmjs.org/hadouken-js-adapter/-/hadouken-js-adapter-0.32.1-alpha.5.tgz",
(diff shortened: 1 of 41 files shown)


pjbroadbent:
Args should be documented within runner.js, much like the PR description above.

Could the util also pass-through any additional arguments whilst we're at it? In case you want to use `--verbose`/`--fail-fast`/etc

tomrobertsOF:
Sorry, the syntax has changed after Phil's comment. There is a description in the file, but basically it no longer uses positional arguments. Will update the description of the PR to match the new syntax

PR #152, opened 2018-09-24
[SERVICE-253] Deregister for Save & Restore

MichaelMCoates:
Deregister now deregisters a window from Save & Restore as well. If an application is deregistered, its child windows will not Save & Restore, even if they aren't deregistered.

- Implement deregister for normal windows
- Handle edge cases when restoring tab groups with a combination of valid windows and child windows of deregistered windows
- Separate LayoutsUI code from demo application/demo window code
- Refactor window object manipulation to a few util functions.

### DIFF
--- a/src/demo/deregisteredApp.ts
+++ b/src/demo/deregisteredApp.ts
@@ -1,66 +1,6 @@
-import {Identity} from 'hadouken-js-adapter';
-import {_Window} from 'hadouken-js-adapter/out/types/src/api/window/window';
-
-import {Layout, LayoutApp} from '../client/types';
-import {positionWindow} from '../provider/workspaces/utils';
-
-export interface Workspace {
-    id: string;
-    layout: Layout;
-}
-
 import * as Layouts from '../client/main';
 
-declare var window: _Window&{forgetMe: (identity: Identity) => void};
-
-let numChildren = 0;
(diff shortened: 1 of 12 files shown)


src/demo/deregisteredApp.ts
```diff
@@ -0,0 +1,66 @@
...
+        } else {
+            await positionWindow(win);
+        }
+    });
+    await Promise.all(openAndPosition);
+    return layoutApp;
+}
+
+// Do not snap to other windows
+Layouts.deregister();
```
pjbroadbent: Can we remove the duplication with `normalApp.ts` here?

I _think_ it should be possible to make this entire file into:
```
import * as Layouts from '../client/main';

export {Workspace, createChild, onAppRes} from './normalApp.ts';    // Or whichever symbols are actually used within the HTML app

// Do not snap to other windows
Layouts.deregister();

MichaelMCoates:
I agree, but once we have deregistration from individual portions of layouts, I think this will make more sense.

src/demo/deregisteredApp.ts
MichaelMCoates: Will do!

MichaelMCoates: Fixed!

PR #154, opened 2018-10-01
'Desktop Model' refactoring

pjbroadbent:
Refactored the service to break-out the state cache from the 'snapanddock' feature to be a service-level abstraction. Re-worked the snap&dock and tabbing features to make use of the new model.

**Summary of changes:**
- Unlike previous `Snap*` model classes, Desktop Model is completely encapsulated, now no longer any references to OpenFin window objects outside of the `DesktopWindow` class.
- Model will track all asynchronous changes that are made by the service. Provides utils for waiting for all changes to complete (`sync()`) and forcing a refresh of the cache (`refresh()`).
- The state cache within the model also tracks which properties were set by the application itself, and which were set by the service. This allows the service to "play nice" with applications.
- Remove duplication of state between snap&dock and tabbing.
- Simplified tabbing implementation, due to having a synchronously-available state cache.
- Simplified tabbing service provider-side API's, removing all optional arguments.
- Reduced number of util files/functions, due to model taking on much of that functionality.
- Ported snap & dock and tabbing to V2 APIs.
- Moved all client-provider communication into a single location (`APIHandler.ts`) - similar to how tabbing worked previously, but now service-wide.
- Added some additional command-line args to test runner - can now optionally skip building of tests before running and run more than one test file at once. Also prevented the double-building of tests that happens on develop branch.

**Known issues and TODO's:** (issues that will be addressed in future stories/PR's)
- Certain tabbing actions are missing some polish compared to develop. This is down to order of operations within the tabbing service.
- Some z-order issues when minimising and maximising tab groups.
- Still inconsistencies in code style between different parts of the codebase. Requires some discussion before cleaning-up.

**Resolved issues** (issues that existed when PR was opened, but are now resolved)
- Tests have gotten more brittle, will all pass individually, but fail fairly consistently when running all tests in one go. Currently under investigation.


src/provider/APIHandler.ts
```diff
@@ -0,0 +1,331 @@
...
+    private async closeTab(tabId: TabIdentifier): Promise<void> {
+        const tab: DesktopWindow|null = model.getWindow(tabId);
+
+        if (tab) {
+            return tab.close();
+        } else {
+            return Promise.reject(`No such tab: ${tabId}`);
+        }
+
+        // const group: DesktopTabGroup|null = tab && tab.getTabGroup();
```
ryanpetersonOF: Comments

src/provider/snapanddock/SnapService.ts
```diff
@@ -211,220 +141,77 @@ export class SnapService {
...
             // The active group should now have been removed (since it is empty)
-            if (this.groups.indexOf(activeGroup) >= 0) {
+            if (groups.indexOf(activeGroup) >= 0) {
                 console.warn(
                     'Expected group to have been removed, but still exists (' + activeGroup.id + ': ' + activeGroup.windows.map(w => w.getId()).join() + ')');
             }
-            // TAB WINDOWS
-        } else if (activeGroup.length === 1) {
-            const currentDragWindowIdentity: WindowIdentity = activeGroup.windows[0].getIdentity();
+        } else if (activeGroup.length === 1 && !activeGroup.windows[0].getTabGroup()) {  // TAB WINDOWS
```
ryanpetersonOF: This is tabbing logic for adding/tabbing two windows together.  Do we want this to live in the SnapService class?

src/demo/LayoutsUI.ts
```diff
@@ -22,6 +21,32 @@ const forgetWindows: Identity[] = [];
 
 window.forgetMe = forgetMe;
 
+export const positionWindow = async (win: WindowState) => {
```
MichaelMCoates: Why was positionWindow moved to here? Was that for separation of demo code from utils?

pjbroadbent: Yes, including provider code within the demo was having some very weird side-effects. The demo was creating it's own Snap&Dock preview window, for example.

Duplication isn't ideal, but we should be keeping provider/client and demo separate. Can see if we can tidy up as part of S&R re-arch work.

src/provider/snapanddock/SnapService.ts
pjbroadbent: Moved implementation to `TabService`

src/provider/APIHandler.ts
pjbroadbent: Removed

pjbroadbent:
Tests are now passing on CI and have completed previous releases, have removed 'do not merge' message.

res/provider/provider.html
```diff
@@ -5,11 +5,13 @@
...
   <script src="main-bundle.js"></script>
-  <title>Layouts Manager Service</title>
+  <title>Layouts Service</title>
 </head>
 
 <body>
-  <h3>"Headless" Layouts Manager Service</h3>
+  <h3>Layouts Service Provider</h3>
+
+  <button onclick="fin.System.showDeveloperTools({uuid: 'layouts-service', name: 'layouts-service'})">Show Debugger</button>
```
tomrobertsOF: These changes should be replicated to `res/test/provider.html` to keep things consistent.

test/provider/fourWindows.test.ts
```diff
@@ -63,7 +63,8 @@ test.failing('inner horizontal resize', async t => {
     t.is(secondRowActualWidth, secondRowExpectedWidth);
 });
 
-test.failing('inner vertical resize', async t => {
+// Test will fail when class is ran individually, but pass when all tests are ran at once. Needs investigation.
+test.skip('inner vertical resize', async t => {
```
tomrobertsOF: Commented about this on the main thread. If this isn't going to be fixed, can we at least find the likely cause and make a bug in JIRA to track it before this is merged?

pjbroadbent: Now resolved

PR #155, opened 2018-10-02
Adds Additional Tests for Tabbing

ryanpetersonOF:
Adds 'TABSET-' as a prefix to the name of all tab set windows.  Format name will be `TABSET-{uuidv4}`


ryanpetersonOF:
Tests now check for inclusion of tab set in window groups.

PR #165, opened 2018-10-10
Save And Restore Update

MichaelMCoates:
- Fix deregister bug in demo windows
- Properly restore child windows and tabbed windows
- Properly reassign the active tab in a tabset when restoring

### DIFF
--- a/src/demo/normalApp.ts
+++ b/src/demo/normalApp.ts
@@ -56,7 +56,7 @@ export async function onAppRes(layoutApp: LayoutApp): Promise<LayoutApp> {
     const openWindows = await ofApp.getChildWindows();
     const openAndPosition = layoutApp.childWindows.map(async (win: WindowState, index: number) => {
         if (!openWindows.some((w: _Window) => w.identity.name === win.name)) {
-            const ofWin = await openChild(win.name, index, win.frame, win.info.url);
+            const ofWin = await openChild(win.name, index, win.frame, win.info.url, win);
         } else {
             await positionWindow(win);
         }
--- a/src/demo/normalApp.ts
+++ b/src/demo/normalApp.ts
@@ -9,6 +9,7 @@ export interface Workspace {
 }
 
 import * as Layouts from '../client/main';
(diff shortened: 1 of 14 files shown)


src/demo/normalApp.ts
```diff
@@ -1,15 +1,15 @@
 import {Identity} from 'hadouken-js-adapter';
 import {_Window} from 'hadouken-js-adapter/out/types/src/api/window/window';
 
-import {Layout, LayoutApp} from '../client/types';
-import {positionWindow} from '../provider/workspaces/utils';
+import {Layout, LayoutApp, WindowState} from '../client/types';
 
 export interface Workspace {
```
ryanpetersonOF: Minor detail, can we move this below the imports?

src/demo/normalApp.ts
```diff
@@ -18,22 +18,39 @@ const launchDir = location.href.slice(0, location.href.lastIndexOf('/'));
...
+            defaultWidth: 250 + 50 * i,
+            defaultLeft: 320 * (i % 3),
+            defaultTop: i > 2 ? 400 : 50,
+            saveWindowState: false,
+            frame,
+            name
+        });
+    }
+
     return win;
```
ryanpetersonOF: You could do `return await fin.Window.create(...)` instead of assigning it to win

tomrobertsOF:
Haven't been able to break it so far. 
It could be a bit confusing from a ux perspective that the basic snap windows don't save/restore, but I'm guessing that's because the layouts-manager deregisters itself.

src/demo/normalApp.ts
```diff
@@ -1,27 +1,80 @@
...
+            await positionWindow(win);
+        }
+    });
+    await Promise.all(openAndPosition);
+    return layoutApp;
+}
+
+// Positions a window when it is restored.
+// Also given to the client to use.
 const positionWindow = async (win: WindowState) => {
```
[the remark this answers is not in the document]
MichaelMCoates: Removed from LayoutsUI.ts

tomrobertsOF:
Nice tidying pass. Looks much cleaner now.

PR #167, opened 2018-10-11
[SERVICE-203] Add drop area calculation 

ryanpetersonOF:
Adds defined areas on the window where tabbing will occur.  The area is set to be the height of the windows tab strip from the windows top edge.


src/provider/tabbing/TabService.ts
```diff
@@ -285,12 +296,30 @@ export class TabService {
...
+                        {x: options.x!, y: options.y!})) {
+                    // standalone window (no tab group)
                     await tabGroup.removeTab(ejectedTab, null);
                     await this.createTabGroupWithTabs([isOverTabWindow.getIdentity(), ejectedTab.getIdentity()]);
+                } else {
+                    // window with tab group, dropped not on tabgroup
+                    const prevHalfSize = ejectedTab.getState().halfSize;
+                    const halfSize = {x: prevHalfSize.x, y: prevHalfSize.y + tabGroup.config.height / 2};
+                    const center = {x: options.x! + halfSize.x, y: options.y! + halfSize.y};
+                    await tabGroup.removeTab(ejectedTab, {center, halfSize});
```
pjbroadbent: Duplication in various parts of the if/else. Is it possible to restructure to remove the duplication, or break out into functions?

src/provider/model/MouseTracker.ts
```diff
@@ -0,0 +1,74 @@
...
+            this.window.onCommit.remove(this.end, this);
+            this.window = this.mouseOffset = null;
+        }
+    }
+
+    /**
+     * Returns the mouse position on screen when a window is being moved. If no window is being moved then we return null.
+     * @returns {PointTopLeft | null} Mouse Position or null.
+     */
+    public getPosition(): PointTopLeft|null {
```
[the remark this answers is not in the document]
ryanpetersonOF: Fixed to use Point instead of PointTopLeft

src/provider/tabbing/TabService.ts
pjbroadbent: I think it should be possible to restructure as follows:

Different scenarios (from reading current code)
* no mouse co-ordinates => Eject at current position
* drag to empty space => Eject tab at position
* drag over untabbed window outside of drop area => Eject tab at position
* drag over existing tab strip => Move tab into existing tab group
* drag over drop area of untabbed window => Create new tab group
* dragged onto same tab group => Do nothing

Re-structured pseudo-code:
```
if (isOverTabGroup && isOverTabWindow === isOverTabGroup.window) {
    // Move tab into existing tab group
} else if (isOverTabGroup && isOverTabGroup !== isOverTabWindow.tabGroup {
	const hasMousePosition: options is Point = (options.x !== undefined && options.y !== undefined);
	const dropArea: Rectangle = { ... };
	
	if (hasMousePosition && RectUtils.isPointInRect(dropArea, options)) {
		// Create new tab group
	} else if (hasMousePosition) {
		// Eject tab, at mouse position
	} else {
		// Eject tab, at it's current position
	}
} else {
	// Tab has been dragged and dropped onto the same tab group, do nothing.
	// This was probably a tab re-ordering operation. This is handled separately.
}
```

tomrobertsOF:
Tests look good. Can you just double-check the linting? Getting errors from `npm run check` when running it locally..

src/provider/model/DesktopTabGroup.ts
```diff
@@ -16,8 +16,20 @@ export class DesktopTabGroup {
...
     public static readonly onDestroyed: Signal1<DesktopTabGroup> = new Signal1();
 
+    /**
+     * Creates a UUIDv4() ID
+     * Sourced from https://stackoverflow.com/questions/105034/create-guid-uuid-in-javascript
+     */
+    private static createTabGroupId(): string {
+        return 'TABSET-' +
+            //@ts-ignore Black Magic
+            ([1e7] + -1e3 + -4e3 + -8e3 + -1e11).replace(/[018]/g, c => (c ^ (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (c / 4)))).toString(16));
```
MichaelMCoates: We've got a function `fin.desktop.getUuid()` that gives a unique ID. 

Also would it be better to just use an ever-incrementing integer?

ryanpetersonOF: I've used fin.desktop.getUuid() now instead.  Re incrementing a number, Id rather just get something back instead of having to keep track / increment a variable each time.

pjbroadbent:
@MichaelMCoates Review comments have been addressed, going to merge now to unblock other work.

If you have any other feedback, will address separately.

PR #174, opened 2018-10-18
[SERVICE-267] Upgrade layouts to runtime 9.61.36.22

tomrobertsOF:
- Bump runtime version for service and all demo apps to `9.61.36.22`
- Bump 'hadouken-js-adapter' to `0.36.1-alpha.1` to match runtime
- Fix issues caused by breaking changes to IAB channels between api versions .34 and .36
- Remove workarounds for runtime issues fixed in .34 and .35


src/provider/APIHandler.ts
```diff
@@ -29,29 +29,13 @@ export class APIHandler {
         });
     }
 
-    public getClientConnection(identity: Identity): ProviderIdentity|null {
```
pjbroadbent: There's also a `getClientConnection` in workspaces/utils.ts which needs removing too. `isClientConnection` is also duplicated, that should use `apiHandler`.

src/provider/workspaces/restore.ts
```diff
@@ -83,10 +83,8 @@ export const restoreLayout = async(payload: Layout, identity: Identity): Promise
     // Push those placeholder windows into tabbedPlaceholdersToWindows object
     // If an app is running, we need to check which of its child windows are open.
     async function createAllPlaceholders(app: LayoutApp) {
-        // We use the v1 version of Application.wrap(...) due to an event-loop bug when
-        // calling the v2 version inside a channel callback. Due for fix in v35
-        const ofApp = fin.desktop.Application.wrap(app.uuid);
-        const isRunning = await p<boolean>(ofApp.isRunning.bind(ofApp))();
```
pjbroadbent: Is the `p` util now unused? If we're (pretty much) fully on the V2 API's now?

src/provider/model/DesktopWindow.ts
```diff
@@ -33,7 +33,7 @@ export interface WindowState extends Rectangle {
     opacity: number;
 }
```
pjbroadbent: There's some additional clean-up that we can do in here now that we have a synchronous `wrap` method.

`window` is currently defined as optional/nullable, due to the ability to create a DesktopWindow from a fin.WindowOptions.

It should now be possible to synchronously wrap the window within the constructor (even though it won't quite exist at that point), and make window non-optional. Can also then clean-up the null checks and `window!`'s.

src/provider/workspaces/restore.ts
tomrobertsOF: Still used in DesktopWindow in `handleFocused`. This needs to stay as a v1 window until the getGroup bug is resolved. Once that is done, we can get rid of it entirely.

src/provider/APIHandler.ts
tomrobertsOF: None of the service connection functions in utils were being used except for a dangling import in `restore.ts`. I've removed the import and deleted all of them.

PR #176, opened 2018-10-24
[SERVICE-130] SnapGroup Validation

tomrobertsOF:
Added implementation for validate group function. Will now split the group if it becomes disjoint.

### DIFF
--- a/src/provider/snapanddock/SnapService.ts
+++ b/src/provider/snapanddock/SnapService.ts
@@ -1,10 +1,7 @@
 import {tabService} from '../main';
 import {DesktopModel} from '../model/DesktopModel';
 import {DesktopSnapGroup, Snappable} from '../model/DesktopSnapGroup';
-import {DesktopTabGroup} from '../model/DesktopTabGroup';
 import {DesktopWindow, eTransformType, Mask, WindowIdentity} from '../model/DesktopWindow';
-import {ApplicationConfigManager} from '../tabbing/components/ApplicationConfigManager';
-import {TabService} from '../tabbing/TabService';
 
 import {EXPLODE_MOVE_SCALE, UNDOCK_MOVE_DISTANCE} from './Config';
 import {eSnapValidity, Resolver, SnapTarget} from './Resolver';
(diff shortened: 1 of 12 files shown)


tomrobertsOF:
@MichaelMCoates I've got what I think should be a fix in place for the tabbing issue. 

Having talked it over a bit, there may have some unintended side-effects on S&R. I think we should leave this open until that PR is merged so we can test more thoroughly before merging this one in.

src/provider/snapanddock/SnapService.ts
```diff
@@ -235,7 +235,8 @@ export class SnapService {
         for (let i = 0; i < windows.length; i++) {
             adjacencyList[i] = [];
             for (let j = 0; j < windows.length; j++) {
-                if (i !== j && RectUtils.distance(windows[i].getState(), windows[j].getState()).border(0)) {
+                const distance = RectUtils.distance(windows[i].getState(), windows[j].getState());
```
pjbroadbent: Unused var `distance`?

src/provider/snapanddock/SnapService.ts
```diff
@@ -264,5 +265,22 @@ export class SnapService {
...
             }
         }
+
+        function isAdjacent(win1: Snappable, win2: Snappable) {
+            const distance = RectUtils.distance(win1.getState(), win2.getState());
+            if (win1.getTabGroup() && win1.getTabGroup() === win2.getTabGroup()) {
+                // Special handling for tab groups. When validating, all windows in a tabgroup are
+                // assumed to be adjacent to avoid weirdness with hidden windows.
+                return true;
+            } else if (win1.getState().hidden || win1.getState().hidden) {
```
pjbroadbent: Should one of these be `win2`?..

tomrobertsOF: Yes

src/provider/snapanddock/SnapService.ts
tomrobertsOF: Yea, got moved to the util and forgot to remove this.

PR #180, opened 2018-10-29
Fix breaking tests

tomrobertsOF:
Some tests were consistently failing only when run in CI. Made some tweaks to resolve.

Need to leave this PR and ensure that it builds properly in CI.


tomrobertsOF:
CI runs seem to be passing now for this PR. Have just kicked one off and will check on it in the morning.

PR #182, opened 2018-11-05
[SERVICE-270] Exclude OpenFin error dialogs from the service

tomrobertsOF:
Added guards in the registration logic to filter out error message windows. Match based on the signature `error-app-[random UUIDv4]`.

### DIFF
--- a/src/provider/model/DesktopModel.ts
+++ b/src/provider/model/DesktopModel.ts
@@ -39,7 +39,13 @@ export class DesktopModel {
         const serviceUUID: string = fin.Application.me.uuid;
 
         // Listen for any new windows created and register them with the service
-        fin.System.addListener('window-created', (evt: WindowEvent<'system', 'window-created'>) => {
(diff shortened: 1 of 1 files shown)


src/provider/model/DesktopModel.ts
```diff
@@ -39,16 +39,22 @@ export class DesktopModel {
...
                 this.registerWindow({name: evt.name, uuid: evt.uuid});
             }
         });
 
         fin.System.getAllWindows().then(apps => {
             apps.forEach((app) => {
-                // Ignore the main service window and all of it's children
-                if (app.uuid !== serviceUUID) {
+                // Ignore openfin error windows, the main service window, and all of it's children
+                if (this.isErrorWindow(app.uuid) && app.uuid !== serviceUUID) {
```
pjbroadbent: Should be `!this.isErrorWindow(app.uuid)`

PR #184, opened 2018-11-05
[SERVICE-243] Client documentation and build process clean-up.

pjbroadbent:
TypeDoc clean-up and validation (as described in story), plus some additional scope creep. Also re-structures the client code to produce more readable output. All changes are non-breaking - the exported interface of the client module is still the same as it was before.

PR also changes the build process, to make the project much easier to debug. Now builds faster, without needing to copy files, with fewer NPM dependencies and package scripts, and using webpack watch behaviours by default.

When running locally, all that is needed to run the demo app is `npm install; npm start`, with no intermediate build step. Local server will also (cleanly) exit when the demo app is closed. Any changes (to both resources and code) can be applied by refreshing the modified page without needing to restart.

This includes some changes to the Jenkinsfile to account for new folder structure and no longer needing to copy files around as part of the build. These changes cannot be tested locally, currently no robust way to validate this until after this PR is merged into develop.

Summary of changes:
- Separated client APIs into feature-specific files, to produce more readable TypeScript API docs.
- Moved client code for intra-service communication into separate files, which are excluded from public-facing TS docs.
- Changed dev build process to no longer copy files or write to disk (by default).
- Changed dev build process to watch file system and re-build any changes.
- Changed dev build process to exit cleanly when demo app is closed.
- When building without debugging, output goes direct to 'dist' directory - replaced existing 'build'/'staging'/'dist' with just 'dist' (assets remain in 'res', not copied to 'dist').
- Added TypeDoc comments to the whole of src/client. This is a first-pass, many parts could do with some improvement.
- Some general clean-up of the demo app. Simplified HTML, renamed some files, etc. Moved windows only used in unit tests from res/demo to res/test.


test/demo/utils/serviceUtils.ts
```diff
@@ -2,7 +2,8 @@
 import {Fin} from 'hadouken-js-adapter';
 import {ChannelClient} from 'hadouken-js-adapter/out/types/src/api/interappbus/channel/client';
 
-import {CHANNEL_NAME} from '../../../src/client/types';
+// import {SERVICE_CHANNEL} from '../../../src/client/internal';
+const SERVICE_CHANNEL = 'of-layouts-service-v1';  // TODO: Remove duplication with internal.ts
```
tomrobertsOF: Can you include a JIRA ref for this TODO?

test/tsconfig.json
```diff
@@ -15,7 +15,7 @@
       "demo/**/*.ts",
       "../src/fin.d.ts",
       "../src/finV2.d.ts",
-      "providerTypes.d.ts"
+      "providerTypes.d.ts", "../src/demo/popup.ts"
```
tomrobertsOF: Minor and nitpicky, but can you put this on its own line

test/demo/utils/serviceUtils.ts
pjbroadbent: I later resolved the issue that was causing this duplication as part of this PR, seems I missed this though. Now fixed, nothing to add to backlog.

test/tsconfig.json
pjbroadbent: Not sure where this came from, it's not necessary. Have tidied the includes slightly.

test/demo/utils/serviceUtils.ts
tomrobertsOF: Cool. Looks good.

test/tsconfig.json
tomrobertsOF: Awesome. Much nicer.

pjbroadbent:
Fixed merge conflict

PR #187, opened 2018-11-06
[SERVICE-270] Exclude OpenFin error dialogs from the service

tomrobertsOF:
Added checks when registering windows to ensure the service ignores any windows with uuid/name matching the pattern `error-app-{UUIDv4}`.


tomrobertsOF:
Michael raised a good point that these will also need to be excluded from generateLayout. Will fix and push up a new commit shortly.

PR #188, opened 2018-11-07
[SERVICE-274] Tab title in tabstrip UI doesn't update if a tab is renamed by an application

ryanpetersonOF:
This PR adds an event listener `tab-properties-updated` which will receive events when the API method `updateTabProperties` is used.  This will resolve the issue previously noted for the tab strip not updating correctly when the API method was called externally.

### DIFF
--- a/test/demo/utils/tabServiceUtils.ts
+++ b/test/demo/utils/tabServiceUtils.ts
@@ -1,9 +1,11 @@
 import {Identity} from 'hadouken-js-adapter';
 
+import {TabAPI} from '../../../src/client/APITypes';
+import {TabProperties, UpdateTabPropertiesPayload} from '../../../src/client/types';
 import {DesktopTabGroup} from '../../../src/provider/model/DesktopTabGroup';
 import {DesktopWindow, WindowIdentity} from '../../../src/provider/model/DesktopWindow';
 
-import {executeJavascriptOnService} from './serviceUtils';
+import {executeJavascriptOnService, sendServiceMessage} from './serviceUtils';
 
 export async function getTabGroupID(identity: Identity): Promise<string|null> {
     function remoteFunc(this: ProviderWindow, identity: WindowIdentity): string|null {
@@ -30,4 +32,12 @@ export async function getTabbedWindows(identity: Identity): Promise<Identity[]> 
         }
     }
     return executeJavascriptOnService<WindowIdentity, Identity[]>(remoteFunc, identity as WindowIdentity);
+}
+
+
+/**
+ * Send a message to the service requesting a provided tabs properties be updated, aka updateTabProperties API call.
+ */
+export async function updateTabProperties(identity: Identity, properties: Partial<TabProperties>) {
+    await sendServiceMessage<UpdateTabPropertiesPayload, void>(TabAPI.UPDATETABPROPERTIES, {window: identity, properties});
 }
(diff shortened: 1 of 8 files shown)


ryanpetersonOF:
@tomrobertsOF Testing has been added

test/provider/updateTabProperties.test.ts
```diff
@@ -0,0 +1,95 @@
...
+test.beforeEach(async () => {
+    wins = await windowInitializer.initWindows(2);
+});
+
+test.afterEach.always(async () => {
+    // Try and close all the windows.  If the window is already closed then it will throw an error which we catch and ignore.
+    await Promise.all(wins.map(win => {
+        try {
+            return win.close();
+        } catch (e) {
```
pjbroadbent: This will only catch any exceptions that happen synchronously within the `close` function.

To suppress any errors due to the window already being closed, will need to add a `.catch` to the `close` promise. Could also catch at the top-level after the `all`, but that could in theory start the next test before all windows have had chance to close.

```
await Promise.all(wins.map(win => {
    return win.close().catch(() => {});
}));
```

test/provider/updateTabProperties.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const newProps = {title: '' + Math.random() * 10, icon: 'http://cdn.openfin.co/favicon.ico'};
+
+    // Update first windows Tab Properties (icon, title);
+    await updateTabProperties(wins[0].identity, newProps);
+
+    function remoteFunc(this: ProviderWindow, identity: WindowIdentity) {
+        const tabGroup = this.model.getWindow(identity as WindowIdentity)!.getTabGroup();
+
+        if (tabGroup) {
```
pjbroadbent: If there's no tab group, then something has gone wrong and this test should fail, right?

Test is currently written in a way where if there is no tab group, the test will pass without any indication that something is broken.

test/demo/utils/tabServiceUtils.ts
```diff
@@ -30,4 +32,12 @@ export async function getTabbedWindows(identity: Identity): Promise<Identity[]>
...
     }
     return executeJavascriptOnService<WindowIdentity, Identity[]>(remoteFunc, identity as WindowIdentity);
+}
+
+
+/**
+ * Send a message to the service requesting a provided tabs properties be updated, aka updateTabProperties API call.
+ */
+export async function updateTabProperties(identity: Identity, properties: Partial<TabProperties>) {
+    await sendServiceMessage<UpdateTabPropertiesPayload, void>(TabAPI.UPDATETABPROPERTIES, {window: identity, properties});
```
pjbroadbent: @tomrobertsOF This is basically duplicating what's in the client. Any ideas for how we can use the client methods instead of re-defining each of the client bindings with the test connection?

Sounds like a tech debt/test framework story.

tomrobertsOF: We've had lots of problems importing the client into tests in the past. At a basic level, the client is meant to run in and OpenFin container, but the tests run in node and have to establish a connection to the runtime, declaring their own fin objects in the process.

There may be a way to do it, but I haven't found it yet.

test/provider/updateTabProperties.test.ts
pjbroadbent: Ignore that, misread the `some` condition at the end of the test.

test/demo/utils/tabServiceUtils.ts
tomrobertsOF: The only possible solution would be to build some sort of dependency injection into the client. That way node could call the methods with its own fin object while leaving the actual client code unaffected.

PR #189, opened 2018-11-07
[SERVICE-273] Service App.json update - remove remote logging and update license key

ryanpetersonOF:
Adds an openfin license key and removes the  remote logging configuration


tomrobertsOF:
Is this change only needed for the provider, or are there other apps that have these settings?

ryanpetersonOF:
@tomrobertsOF It says right in the title.  This change is for the service app.json only...  Any other app manifests would only be used in the instance of the demo being run.  But as you will find, these settings only exist in the service app.json and no others.

PR #194, opened 2018-11-09
[SERIVCE-135] Respect window size constraints when snapping

tomrobertsOF:
Summary of changes: 
 - Updated the model's windowState interface to include resizable and resizeRegion.
 - Added public method on DesktopWindow to allow getting constraints data in a more manageable format.
 - Added checks to projector that block resizing if it would violate these constraints, or min/max height/width.

### DIFF
--- a/src/provider/model/DesktopSnapGroup.ts
+++ b/src/provider/model/DesktopSnapGroup.ts
@@ -30,12 +30,19 @@ interface TabState {
     previousState: TabData;
 }
 
+export interface ResizeConstraints {
+    resizable: boolean;
+    min: number;
+    max: number;
+}
+
 export interface Snappable {
     getId(): string;
     getIdentity(): WindowIdentity;
     getState(): WindowState;
     getTabGroup(): DesktopTabGroup|null;
     getSnapGroup(): DesktopSnapGroup;
+    getResizeConstraints(orientation: 'x' | 'y'): ResizeConstraints;
 
     // tslint:disable-next-line:no-any
     applyOverride(property: keyof WindowState, value: any): Promise<void>;
--- a/src/provider/model/DesktopSnapGroup.ts
+++ b/src/provider/model/DesktopSnapGroup.ts
@@ -42,7 +42,7 @@ export interface Snappable {
     getState(): WindowState;
     getTabGroup(): DesktopTabGroup|null;
     getSnapGroup(): DesktopSnapGroup;
-    getResizeConstraints(orientation: 'x' | 'y'): ResizeConstraints;
+    getResizeConstraints(orientation: 'x'|'y'): ResizeConstraints;
 
     // tslint:disable-next-line:no-any
     applyOverride(property: keyof WindowState, value: any): Promise<void>;
--- a/src/provider/model/DesktopSnapGroup.ts
+++ b/src/provider/model/DesktopSnapGroup.ts
@@ -30,19 +30,12 @@ interface TabState {
     previousState: TabData;
 }
 
-export interface ResizeConstraints {
-    resizable: boolean;
-    min: number;
-    max: number;
-}
-
 export interface Snappable {
     getId(): string;
     getIdentity(): WindowIdentity;
     getState(): WindowState;
     getTabGroup(): DesktopTabGroup|null;
     getSnapGroup(): DesktopSnapGroup;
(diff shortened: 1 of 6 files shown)


src/provider/model/DesktopWindow.ts
```diff
@@ -30,11 +30,17 @@ export interface WindowState extends Rectangle {
     maxWidth: number;
     minHeight: number;
     maxHeight: number;
+    resizable: boolean;
+    resizeRegion: WindowResizeRegion;
```
pjbroadbent: Rather than copying these "min/max/resize" values straight from the runtime state, can we convert them to a format that works better for what we're doing in the service? Like how we do with bounds => center/halfSize? We're kinda doing it already with the `getResizeConstraints` method, but would make more sense IMO to do this at the layouts/runtime boundary rather than on every transform.

Replacing these 6 properties with `constraints: Point<ResizeContstraint>` would certainly tidy things up a bit. Should ideally define `ResizeContstaint` in a way that supports SERVICE-278 without us needing to tweak this again.

src/provider/model/DesktopSnapGroup.ts
```diff
@@ -30,12 +30,19 @@ interface TabState {
...
+    max: number;
+}
+
 export interface Snappable {
     getId(): string;
     getIdentity(): WindowIdentity;
     getState(): WindowState;
     getTabGroup(): DesktopTabGroup|null;
     getSnapGroup(): DesktopSnapGroup;
+    getResizeConstraints(orientation: 'x'|'y'): ResizeConstraints;
```
pjbroadbent: There's an `Orientation` type in Resolver.ts

src/provider/snapanddock/Projector.ts
```diff
@@ -112,7 +112,13 @@ export class Projector {
                                 Math.abs((activeState.center[border.opposite] + activeState.halfSize[border.opposite]) - border.max) < ANCHOR_DISTANCE;
 
                             if (snapToMin && snapToMax) {
-                                halfSize[border.opposite] = (border.max - border.min) / 2;
+                                const {minSize, maxSize, resizableMin, resizableMax} = activeState.resizeConstraints[border.opposite];
+                                const targetBorderLength = (border.max - border.min);
+
+                                // Only resize if it would not violate any constraints
+                                if ((resizableMin || resizableMax) && targetBorderLength > minSize && targetBorderLength < maxSize) {
```
pjbroadbent: `>=`/`<=`?

src/provider/model/DesktopWindow.ts
```diff
@@ -26,15 +26,19 @@ export interface WindowState extends Rectangle {
...
 
     opacity: number;
     frameEnabled: boolean;  // If window will respond to move/resize events. Corresponds to enable/disableFrame, not WindowOptions.frame.
 }
 
+export interface ResizeConstraint {
+    resizableMin: boolean;
+    resizableMax: boolean;
+    minSize: number;
+    maxSize: number;
```
pjbroadbent: Can you align the names on these? Either both `minX` or `xMin`?

src/provider/model/DesktopWindow.ts
```diff
@@ -334,10 +351,10 @@ export class DesktopWindow extends DesktopEntity implements Snappable {
             icon: '',
             title: '',
             showTaskbarIcon: true,
-            minWidth: -1,
-            maxWidth: -1,
-            minHeight: 0,
-            maxHeight: 0,
+            resizeConstraints: {
+                x: {minSize: 0, maxSize: Number.MAX_SAFE_INTEGER, resizableMin: true, resizableMax: false},
+                y: {minSize: 0, maxSize: Number.MAX_SAFE_INTEGER, resizableMin: true, resizableMax: false}
```
pjbroadbent: Should `resizableMin` and `resiableMax` have different defaults?

tomrobertsOF: No. Not sure where that came from.

src/provider/snapanddock/Projector.ts
tomrobertsOF: Yea, good point.

src/provider/model/DesktopWindow.ts
tomrobertsOF: Let's discuss offline.

PR #196, opened 2018-11-23
[SERVICE-264] Add Visual Previews for window-to-window Tab "drop area" actions

ryanpetersonOF:
This adds visual feedback, similar to snapping feedback, for tabbing interactions.  There are refactor works with the View (previously SnapView) and Preview (previously SnapPreview) classes.

It also sets up the future implementation for tab dragging visual feedback, in a separate story.  

Testing is coming shortly - I've made this PR for feedback on the initial changes.  Please dont merge before testing has been added.


src/provider/WindowHandler.ts
```diff
@@ -0,0 +1,113 @@
...
+    private onGroupTransform(activeGroup: DesktopSnapGroup, type: Mask<eTransformType>) {
+        this.view.update(activeGroup, this.getTarget(activeGroup));
+    }
+
+    private onGroupCommit(activeGroup: DesktopSnapGroup) {
+        const target = this.getTarget(activeGroup);
+
+        if (target) {
+            if (target.type === eTargetType.TAB) {
+                tabService.tabDroppedWindow(activeGroup.windows[0] as DesktopWindow);
```
tomrobertsOF: This seems strange to me. Wouldn't this then go and do a load of target selection logic again? Since we've already done most of the work, couldn't we have the tabService accept a target object and use that to create the tabset directly? (something like `applyTabTarget(target)`)

src/provider/WindowHandler.ts
```diff
@@ -0,0 +1,113 @@
...
+     */
+    activeWindow: Snappable;
+
+    /**
+     * The offset that will be applied to the active window, in order to correctly align it with this target.
+     */
+    offset: Point;
+
+    /**
+     * If 'activeWindow' should be resized as part of this snap, it's new halfSize will be specified here. This only
```
tomrobertsOF: Nitpicky, but this comment seems very snap specific, while the class seems to be (mostly) agnostic to which operation was being performed.

src/provider/main.ts
```diff
@@ -36,6 +37,7 @@ type Stringified<T> = {
 
 export async function main() {
     model = window.model = new DesktopModel();
+    windowHandler = new WindowHandler(model);
```
tomrobertsOF: is there any specific reason why we don't mount this on the provider window like the others? I don't mind either way, just want to be sure it's an intentional decision.

src/provider/View.ts
```diff
@@ -57,10 +69,30 @@ export class SnapView {
                     window.applyOverride('opacity', 0.8);
                 });
             } else {
+                // group.windows[0].applyOverride('alwaysOnTop', false);
```
pjbroadbent: Remove commented-out code

src/provider/tabbing/TabService.ts
```diff
@@ -309,11 +310,65 @@ export class TabService {
...
+
+        if (targetWindow && dropArea) {
+            const isTargetTabbed = targetWindow.getTabGroup();
+            const targetWindowState = isTargetTabbed && isTargetTabbed.window.getState() || targetWindow.getState();
+            const activeWindowState = activeGroup.windows[0].getState();
+
+            // Calculate offset between the target window & the active Window
+            const offset: Point = PointUtils.difference(
+                RectUtils.getCorner('top-left', activeWindowState.center, activeWindowState.halfSize),
+                RectUtils.getCorner('top-left', targetWindowState.center, targetWindowState.halfSize));
```
pjbroadbent: This is why `offset` doesn't really make much sense in the tabbing context. We have to convert the drop area rect into an offset, just to convert it back to a rect again. Converting the snap service to use bounds over offset is also not straight-forward ('offset' propagates right down into `dockToGroup`).

Separate snap/tab targets may be cleaner..?

src/provider/View.ts
```diff
@@ -57,10 +69,30 @@ export class SnapView {
...
                     window.resetOverride('opacity');
                 });
             }
         }
     }
+
+    /**
+     * Applys alwaysOnTop to the primary window of a desktop snap group.  Required to keep the preview window in proper z-index order under the active window.
+     * @param {DesktopSnapGroup} group The activeGroup being dragged by the user.
+     * @param {boolean} applyOnTop Apply alwaysOnTop or not.
```
pjbroadbent: Remove types from `@param` tags - typedoc will grab them from the code. Copying them into the code only adds overhead - the `group` type is already wrong, as it doesn't capture the `|null`.

src/provider/WindowHandler.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Fetches the appropriate target from different services.
+     * @param {DesktopSnapGroup} activeGroup The active group being moved by the user.
+     */
+    private getTarget(activeGroup: DesktopSnapGroup): Target|null {
+        const groups: ReadonlyArray<DesktopSnapGroup> = this.model.getSnapGroups();
+        const snapTarget: Target|null = snapService.resolver.getSnapTarget(groups, activeGroup);
+        const tabTarget = (activeGroup.windows.length === 1 || activeGroup.windows[0].getTabGroup()) && tabService.getTarget(activeGroup);
```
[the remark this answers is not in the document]
ryanpetersonOF: fixed

src/provider/View.ts
ryanpetersonOF: fixed

tomrobertsOF:
One extra thing to report. It looks like the preview windows are being shown a split second before their state is updated to reflect the new target. I think this was always an issue, but it is particularly noticeable now that there are brightly coloured previews. 

If you take a custom tab strip window and try to tab it to a default, then move it into a tabbing position, you will see the preview window appear with red stripes for a small fraction of a second before showing the expected preview. Could we maybe move the ordering of the show around so that the position/colour of the preview is correct before calling `show()`?

tomrobertsOF:
Also, can you add some tests for the new previews. There are some examples in `test/provider/previewWindow.test.ts` that should be easy enough to adapt for the tab previews.

ryanpetersonOF:
@tomrobertsOF Re: Preview Window State - Known issue.  I've had a play around, seeing if it could be fixed during this ticket.  The issue is that the window is not rendered while it is hidden so switching the logic around as suggested won't help here.  Further investigation/rework will be needed for it.  Since it's not fully in the scope of this story we have created SERVICE-287 for it.

Re: Testing - This is coming shortly as the PR description states.  Working on it at this very moment.  Will let you know once its in.

tomrobertsOF:
@ryanpetersonOF sorry, didn't see that in the description (should read more carefully). Re the preview window state: that's fine. I figured if it was a simple change of order then may as well do it, otherwise fine to leave it.

src/provider/main.ts
ryanpetersonOF: No major reason.  There's not much you can really do to interact with it though.  It just needed to be high enough level to be in main so its there. If you can think of any need to make it on the window I can change it.

src/provider/WindowHandler.ts
ryanpetersonOF: Agreed. This entire section is currently being evaluated for the proper terminology (comments and keys).  It's likely to change again before a merge.

src/provider/main.ts
tomrobertsOF: Nah that's fine. As I said, just wanted it to be an intentional decision. Since everything on it is private anyway, makes sense to just leave it internal.

src/provider/WindowHandler.ts
tomrobertsOF: Fair enough. I'll keep an eye out.

test/demo/previewWindow.test.ts
```diff
@@ -0,0 +1,119 @@
...
+
+        await dragSideToSide(windows[1], opposite(side), windows[0], side, {x: 5, y: 5}, false);
+
+        const previewBounds = await getBounds(previewWin);
+        robot.mouseToggle('up');
+
+        t.is(windowBounds[1].width, previewBounds.width);
+        t.is(windowBounds[1].height, previewBounds.height);
+
+        await assertAdjacent(t, windows[0], previewWin);
```
tomrobertsOF: Would it not be better to check this before dropping the window (i.e. between lines 38 and 39)?

test/demo/previewWindow.test.ts
```diff
@@ -0,0 +1,119 @@
...
+            'top-left');
+
+        dimension === 'height' ? await dragWindowAndHover(windows[1], windowBounds[0].right, windowBounds[0].top) :
+                                 await dragWindowAndHover(windows[1], windowBounds[0].left, windowBounds[0].bottom);
+
+        const previewBounds = await getBounds(previewWin);
+
+        robot.mouseToggle('up');
+
+        dimension === 'height' ? t.is(previewBounds.height, windowBounds[0].height) : t.is(previewBounds.width, windowBounds[0].width);
```
tomrobertsOF: This ternary statement isn't really needed since dimension takes a subset of the keys of previewBounds. 
You could just do: `t.is(previewBounds[dimension], windowBounds[0][dimension]`

src/provider/tabbing/TabService.ts
```diff
@@ -1,16 +1,24 @@
...
 import {DesktopTabGroup} from '../model/DesktopTabGroup';
 import {DesktopWindow, WindowState} from '../model/DesktopWindow';
 import {Rectangle, RectUtils} from '../snapanddock/utils/RectUtils';
-
+import {eTargetType, TargetBase} from '../WindowHandler';
 import {ApplicationConfigManager} from './components/ApplicationConfigManager';
 import {DragWindowManager} from './DragWindowManager';
 
 
+export interface TabTarget extends TargetBase {
```
pjbroadbent: Add some doc here? Similar to `SnapTarget`/`TargetBase`.

src/provider/tabbing/TabService.ts
```diff
@@ -309,11 +317,70 @@ export class TabService {
...
+            // Check if the target and active window have same tab config.
+            const valid: boolean = this.applicationConfigManager.compareConfigBetweenApplications(
+                isTargetTabbed ? isTargetTabbed.config : targetWindow.getIdentity().uuid, activeGroup.windows[0].getIdentity().uuid);
+
+            return {
+                type: eTargetType.TAB,
+                activeWindow: activeGroup.windows[0],
+                group: targetWindow.getSnapGroup(),
+                dropArea: this.getWindowDropArea(targetWindow),
+                valid
```
pjbroadbent: Is there anything else in here that would be worth adding to the 'target' object? Things that otherwise need re-calculating within the "applyTarget" call.

Target objects are always very short-lived objects, so ok to also include any volatile/temporal data if it would help simply the "apply" parts.

test/demo/previewWindow.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        const previewBounds = await getBounds(previewWin);
+
+        robot.mouseToggle('up');
+
+        t.is(previewBounds.width, windowBounds[0].width);
+        t.is(previewBounds.top, windowBounds[0].top);
+        t.is(previewBounds.left, windowBounds[0].left);
+        t.is(previewBounds.right, windowBounds[0].right);
+        t.is(previewBounds.height, 60);
```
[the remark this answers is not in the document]
tomrobertsOF: You could just use `t.deepEqual` to compare the rects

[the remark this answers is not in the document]
tomrobertsOF: (with the 60 substituted in of course)

src/provider/tabbing/TabService.ts
ryanpetersonOF: This will be better determined in the next ticket for drag & drop, when we will change over the main accepting method to use a target.

src/provider/WindowHandler.ts
```diff
@@ -73,9 +73,10 @@ export class WindowHandler {
 
         if (target) {
             if (target.type === eTargetType.TAB) {
+                // TODO: Change this to accept a target
```
tomrobertsOF: Can you put a JIRA ref with this TODO.

tomrobertsOF: Thanks!

PR #207, opened 2018-12-17
[SERVICE-200] Make tab groups snappable

pjbroadbent:
Can now dock tab groups with other windows/tab groups.

Snap & Dock now considers tab groups to be single entities, meaning they can be snapped together in the same way as regular windows. This requires changes at all levels of the model, and requires separating "windows" from "entities".

Changes include some general refactoring. Re-defined the meaning of `DesktopEntity`, renaming "snappables" to "entities". Entities are anything that should behave as a single window, i.e. `DesktopWindow`s and `DesktopTabGroup`s. Have also aligned codebase around getters rather than the previous mix of getters and "getX" functions.


src/provider/model/DesktopSnapGroup.ts
```diff
@@ -254,9 +250,7 @@ export class DesktopSnapGroup {
         if (type === eTransformType.MOVE) {
             // When a grouped window is moved, all windows in the group will fire a move event.
             // We want to filter these to ensure the group only fires onTransform once
-            if (window === this.rootWindow) {
-                this.onTransform.emit(this, type);
-            }
+            this.onTransform.emit(this, type);
```
MichaelMCoates: How are we handling this filtering now?

pjbroadbent: It's at the window level - only user-initiated events trigger the window commit/modified signals to fire, so we don't need to check here as we know only one of the windows within the group will ever trigger this callback.

This has been the case for a while. If I remember correctly, if you drag a group around on develop, only one of the snapped windows will cause the opacity effect to trigger.

PR #213, opened 2019-01-02
[SERVICE-289] Add config schema

tomrobertsOF:
Initial attempt at a config schema built from Phil's POC/prototype work.

Run `npm run generate` to create the output .d.ts files. They are currently put in `/dist/config-schema`, but that is up for discussion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "start": "node server.js",
     "launch": "node launch.js",
     "zip": "node scripts/createProviderZip.js",
-    "generate": "npm run gen:config && npm run gen:scope",
+    "generate": "mkdir -p dist/config-schema && npm run gen:config && npm run gen:scope",
     "gen:config": "json2ts --input res/config-schema/layouts-config.schema.json -o dist/config-schema/layouts-config.d.ts --cwd res/config-schema",
     "gen:scope": "json2ts --input res/config-schema/scope.schema.json -o dist/config-schema/scope.d.ts --cwd res/config-schema"
   },
--- a/package.json
+++ b/package.json
@@ -21,7 +21,10 @@
     "watch": "webpack --watch --mode development",
     "start": "node server.js",
     "launch": "node launch.js",
-    "zip": "node scripts/createProviderZip.js"
+    "zip": "node scripts/createProviderZip.js",
+    "generate": "npm run gen:config && npm run gen:scope",
+    "gen:config": "json2ts --input res/config-schema/layouts-config.schema.json -o dist/config-schema/layouts-config.d.ts --cwd res/config-schema",
+    "gen:scope": "json2ts --input res/config-schema/scope.schema.json -o dist/config-schema/scope.d.ts --cwd res/config-schema"
   },
   "keywords": [],
   "license": "Apache-2.0",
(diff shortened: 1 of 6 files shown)


res/config-schema/common.schema.json
```diff
@@ -0,0 +1,27 @@
...
+    "definitions": {
+        "pattern": {
+            "title": "RegEx",
+            "description": "JSON representation of a regular expression.",
+            "type": "object",
+            "required": [
+                "regex"
+            ],
+            "properties": {
+                "expression": {
```
pjbroadbent: Can we make `expression` required? Other fields are fine as optional.

res/config-schema/layouts-config.schema.json
```diff
@@ -0,0 +1,88 @@
...
+                },
+                "features": {
+                    "properties": {
+                        "dock": {
+                            "type": "boolean",
+                            "description": "Disables the docking functionality of the service. Windows will still snap side-by-side, but will not be grouped as a result."
+                        },
+                        "tab": {
+                            "type": "boolean",
+                            "description": "Disables the ability to tab windows by drag-and-drop. Windows can still be tabbed programmatically using the layouts client api."
```
pjbroadbent: "Enables", rather than "Disables"?.. I get that you'd only ever specify this to disable a function, but a value of `true` here means "enable" rather than "disable".

Could also go with "controls"/"toggles"/etc...

res/config-schema/scope.schema.json
```diff
@@ -0,0 +1,78 @@
...
+            },
+            "required": [
+                "level",
+                "uuid",
+                "name"
+            ],
+            "additionalProperties": false
+        },
+        "scope-other": {
+            "title": "OtherScope",
```
pjbroadbent: Add a description

res/config-schema/layouts-config.schema.json
```diff
@@ -0,0 +1,88 @@
...
+            "title": "ConfigurationObject",
+            "description": "An instance of Layouts Service configuration.",
+            "type": "object",
+            "properties": {
+                "enabled": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Enables/disables all Layouts Service behaviour for the given scope."
+                },
+                "features": {
```
pjbroadbent: Can you add a description for `features`? Can be brief.

package.json
```diff
@@ -21,7 +21,10 @@
     "watch": "webpack --watch --mode development",
     "start": "node server.js",
     "launch": "node launch.js",
-    "zip": "node scripts/createProviderZip.js"
+    "zip": "node scripts/createProviderZip.js",
+    "generate": "mkdirp dist/config-schema && npm run gen:config && npm run gen:scope",
+    "gen:config": "json2ts --input res/config-schema/layouts-config.schema.json -o dist/config-schema/layouts-config.d.ts --cwd res/config-schema",
+    "gen:scope": "json2ts --input res/config-schema/scope.schema.json -o dist/config-schema/scope.d.ts --cwd res/config-schema"
```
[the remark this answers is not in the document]
tomrobertsOF: This seems like a reasonable suggestion, but probably outside the scope of this story.

res/config-schema/common.schema.json
tomrobertsOF: That was the intention, but I renamed `regex` to `expression` but forgot to update the required array.

res/config-schema/layouts-config.schema.json
tomrobertsOF: Went with "toggles". Seems most descriptive.

package.json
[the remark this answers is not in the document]
pjbroadbent: True, will find an implementation story to add it to.

PR #219, opened 2019-01-08
[SERVICE-293] Fix flicker when restoring snapped tab groups

tomrobertsOF:
Added some guards and tweaked order of operations on events (mostly in DesktopWindow) to prevent looping/cascading events.

This PR involves changes to some fairly low-level service logic. Automatic testing passes and I have done manual testing where possible, but it would be good to have a few other people try to break it.


src/provider/model/DesktopTabGroup.ts
```diff
@@ -223,12 +223,9 @@ export class DesktopTabGroup implements DesktopEntity {
...
     public async minimize(): Promise<void> {
-        const minWins = this._tabs.map(tab => {
-            return tab.applyProperties({state: 'minimized'});
-        });
-        const group = this._window.applyProperties({state: 'minimized'});
-
-        return Promise.all([minWins, group]).then(() => {});
+        // Only minimize the tabstrip and active tab since minimizing hidden windows causes issues.
+        // This may cause problems if switching tabs while minimized, but that would require a questionable custom tabstrip.
+        return Promise.all([this._window.applyProperties({state: 'minimized'}), this.activeTab.applyProperties({state: 'minimized'})]).then(() => {});
```
ryanpetersonOF: Do you need the `.then(...)` on the end?

tomrobertsOF: Kind of. The function (and ones that depend on it) is meant to return 'promise<void>', but Promise.all returns 'Promise<void[]>'. They are basically the same thing, but typescript doesn't like it.

tomrobertsOF: I guess I could just await it and not have a return

test/provider/utils/assertions.ts
```diff
@@ -188,4 +199,19 @@ export async function assertNoOverlap(t: TestContext, windows: Window[]) {
...
+
+export async function assertAllMinimizedOrHidden(t: TestContext, windows: Window[]) {
+    return Promise.all(windows.map(async win => {
+        const showing = await win.isShowing();
+        const state = await win.getState();
+        t.true(state === 'minimized' || !showing);
+    }));
+}
+
+export async function assertAllRestored(t: TestContext, windows: Window[]) {
```
pjbroadbent: Naming could confuse this with save/restore.

tomrobertsOF: Agreed, but what would you suggest as a better name? Un-minimizing a window is known as restoring both in Openfin and generally. This is triggered by calling `window.restore()` and relies on the `window-restored` event.

test/demo/utils/workspacesUtils.ts
```diff
@@ -78,4 +79,16 @@ export function createTabTests(numApps: number, children: number): {apps: AppIni
...
+export async function closeAllPreviews(): Promise<void> {
+    const serviceApp = fin.Application.wrapSync(SERVICE_IDENTITY);
+    const children = await serviceApp.getChildWindows();
+    const actions: Promise<void>[] = [];
+    for (const child of children) {
+        if (child.identity.name!.startsWith('Placeholder-')) {
+            actions.push(child.close());
+        }
+    }
+    return Promise.all(actions).then(() => {});
```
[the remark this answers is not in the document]
tomrobertsOF: I can add some logic that first checks if one is hanging around and fails the test before closing it?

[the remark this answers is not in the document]
tomrobertsOF: Added

test/provider/utils/assertions.ts
tomrobertsOF: I changed it to `assertAllNormalState`

test/demo/utils/workspacesUtils.ts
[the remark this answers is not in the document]
pjbroadbent: We should do that, but sounds like even though this is fairly rare, there's enough placeholder windows created during a full run for something like that to interfere with CI.

Will add something to the backlog to both add the assert, and also find & fix the underlying issue. Actual fix is out of scope of this story.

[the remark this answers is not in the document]
pjbroadbent: Logged as SERVICE-310.

test/provider/utils/assertions.ts
pjbroadbent: Good question :)

Something along the lines of `assertWindowStateNormal`? Can't think of a good name, but that's my least-worst...

test/demo/utils/workspacesUtils.ts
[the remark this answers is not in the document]
pjbroadbent: Seems I didn't refresh before replying! Everything passed on CI with the extra assertions so looks good to go. Will close the ticket I created.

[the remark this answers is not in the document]
MichaelMCoates: I think we might want to re-open that ticket. The placeholder bug has been annoying me for a bit.

PR #220, opened 2019-01-08
[SERVICE-279] Add Visual Previews for tab-to-window Tab "drop area" actions

ryanpetersonOF:
Part 2 of this story.  Dragging tabs around will now produce visual feedback.  Some internal changes regarding Tab/Eject targets are also included.

### DIFF
--- a/src/provider/APIHandler.ts
+++ b/src/provider/APIHandler.ts
@@ -300,6 +300,6 @@ export class APIHandler {
         }
 
         tabService.dragWindowManager.hideWindow();
-        tabService.ejectTab(payload.window, {x: payload.event.screenX, y: payload.event.screenY});
+        tabService.ejectTab(payload.window);
     }
 }
--- a/src/provider/View.ts
+++ b/src/provider/View.ts
@@ -1,6 +1,6 @@
 import {DesktopSnapGroup, Snappable} from './model/DesktopSnapGroup';
 import {Preview, PreviewableTarget} from './Preview';
-import {Target, eTargetType} from './WindowHandler';
+import {eTargetType, Target} from './WindowHandler';
 
 export class View {
     private activeGroup: DesktopSnapGroup|null;  // The group being moved
@@ -24,7 +24,7 @@ export class View {
      * original opacities once the active/target group(s) change or get reset.
      */
     public update(activeGroup: DesktopSnapGroup|null, target: Target|null): void {
-        if(target && target.type === eTargetType.EJECT){
+        if (target && target.type === eTargetType.EJECT) {
             activeGroup = target = null;
         }
 
@@ -47,7 +47,8 @@ export class View {
 
         // Detect change of target group
         if ((this.target && this.isPreviewable(this.target) && this.target.group) !== (target && target.group)) {
-            const targetGroup = this.target && this.isPreviewable(this.target) ? this.target.group : null;// && this.target.type !== eTargetType.EJECT ? this.target.group : null;
+            const targetGroup = this.target && this.isPreviewable(this.target) ? this.target.group :
+                                                                                 null;  // && this.target.type !== eTargetType.EJECT ? this.target.group : null;
 
             // Reset alwaysOnTop override, as our activeGroup window is now in the target group.
             this.setAlwaysOnTop(targetGroup, false);
--- a/src/provider/View.ts
+++ b/src/provider/View.ts
@@ -1,6 +1,6 @@
 import {DesktopSnapGroup, Snappable} from './model/DesktopSnapGroup';
-import {Preview} from './Preview';
-import {Target} from './WindowHandler';
+import {Preview, PreviewableTarget} from './Preview';
+import {Target, eTargetType} from './WindowHandler';
 
 export class View {
     private activeGroup: DesktopSnapGroup|null;  // The group being moved
@@ -24,6 +24,10 @@ export class View {
      * original opacities once the active/target group(s) change or get reset.
      */
     public update(activeGroup: DesktopSnapGroup|null, target: Target|null): void {
+        if(target && target.type === eTargetType.EJECT){
+            activeGroup = target = null;
+        }
+
         // Handle change of active group
         if (activeGroup !== this.activeGroup) {
             // Reset active window always on top property.
@@ -42,8 +46,8 @@ export class View {
         }
 
         // Detect change of target group
-        if ((this.target && this.target.group) !== (target && target.group)) {
-            const targetGroup = this.target && this.target.group;
+        if ((this.target && this.isPreviewable(this.target) && this.target.group) !== (target && target.group)) {
+            const targetGroup = this.target && this.isPreviewable(this.target) ? this.target.group : null;// && this.target.type !== eTargetType.EJECT ? this.target.group : null;
 
             // Reset alwaysOnTop override, as our activeGroup window is now in the target group.
             this.setAlwaysOnTop(targetGroup, false);
@@ -57,11 +61,15 @@ export class View {
 
         // Update preview window
         this.target = target;
-        if (activeGroup && target) {
+        if (activeGroup && target && this.isPreviewable(target)) {
             this.preview.show(target);
         } else {
             this.preview.hide();
         }
+    }
+
+    private isPreviewable(target: Target|null): target is PreviewableTarget {
+        return !!target && (target.type === eTargetType.SNAP || target.type === eTargetType.TAB);
     }
 
     private setGroupOpacity(group: DesktopSnapGroup|null, transparent: boolean): void {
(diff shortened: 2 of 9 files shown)


src/provider/View.ts
```diff
@@ -1,100 +1,109 @@
...
+            this.setAlwaysOnTop(this.activeGroup, true);
+
+            // Apply opacity to active group.
+            this.setGroupOpacity(this.activeGroup, true);
+        }
+
+        // Detect change of target group
+        if ((this.target && this.isPreviewable(this.target) && this.target.group) !== (target && target.group)) {
+            const targetGroup = this.target && this.isPreviewable(this.target) ? this.target.group :
+                                                                                 null;  // && this.target.type !== eTargetType.EJECT ? this.target.group : null;
```
pjbroadbent: Clean-up commented-out code

src/provider/model/MouseTracker.ts
```diff
@@ -15,9 +15,13 @@ export class MouseTracker {
      */
     private _mouseOffset: Point|null = null;
 
+    private knownPosition: Point|null = null;
```
pjbroadbent: `_knownPosition`

src/provider/model/MouseTracker.ts
```diff
@@ -84,4 +99,24 @@ export class MouseTracker {
             this._window = this._mouseOffset = null;
         }
     }
+
+    /**
+     * Returns the mouse position on screen when a window is being moved. If no window is being moved then we return null.
+     * @returns {Point | null} Mouse Position or null.
+     */
+    public getPosition(): Point|null {
```
pjbroadbent: Arrange methods public -> protected -> private

src/provider/tabbing/DragWindowManager.ts
```diff
@@ -1,29 +1,30 @@
...
+import {DesktopModel} from '../model/DesktopModel';
+import {DesktopWindow} from '../model/DesktopWindow';
+import {Signal0, Signal2} from '../Signal';
 
 /**
  * Handles the Drag Window which appears when API drag and drop is initialized.
  */
 export class DragWindowManager {
+    public static readonly onDragOver: Signal2<DesktopWindow, Point> = new Signal2();
+    public static readonly onDragDrop: Signal0 = new Signal0();
```
pjbroadbent: Document these signals, as the non-static signals in `DesktopSnapGroup`. Particularly the "Arguments: " line.

src/provider/tabbing/DragWindowManager.ts
```diff
@@ -1,29 +1,30 @@
...
+
     // Multiple definitions of setTimeout/clearTimeout, and not possible to point TSC at the correct (non-Node) definition
     private _hideTimeout: number|NodeJS.Timer;
 
     private _window!: Window;
 
-    constructor() {
+    private sourceWindow: DesktopWindow|null;
+    private model: DesktopModel;
```
pjbroadbent: `_sourceWindow`/`_model`

src/provider/tabbing/TabService.ts
```diff
@@ -2,11 +2,11 @@ import {Point} from 'hadouken-js-adapter/out/types/src/api/system/point';
 import {ApplicationUIConfig, TabGroup, TabGroupDimensions, WindowIdentity} from '../../client/types';
 import {DesktopEntity} from '../model/DesktopEntity';
 import {DesktopModel} from '../model/DesktopModel';
-import {DesktopSnapGroup} from '../model/DesktopSnapGroup';
+import {DesktopSnapGroup, Snappable} from '../model/DesktopSnapGroup';
 import {DesktopTabGroup} from '../model/DesktopTabGroup';
 import {DesktopWindow, EntityState} from '../model/DesktopWindow';
 import {Rectangle, RectUtils} from '../snapanddock/utils/RectUtils';
-import {eTargetType, TargetBase} from '../WindowHandler';
+import {eTargetType,  TargetBase} from '../WindowHandler';
```
pjbroadbent: Need to run the linter?..

src/provider/tabbing/TabService.ts
```diff
@@ -235,7 +247,7 @@ export class TabService {
      * @param ejectPosition The current position of the mouse (if known). Ejected tab will be moved to this location, and tabbed with any other (compatible)
      * window/tabgroup at this position.
      */
-    public async ejectTab(tab: WindowIdentity, ejectPosition?: Point): Promise<void> {
+    public async ejectTab(tab: WindowIdentity): Promise<void> {
```
pjbroadbent: Is this still used/required? Should it be an ejectTarget now?

Either way needs updating in terms of doc and maybe visibility & naming.

src/provider/model/MouseTracker.ts
ryanpetersonOF: This was in order on my branch. I think merge goof'd it.  Fixed next push

src/provider/tabbing/TabService.ts
ryanpetersonOF: Was able to remove this completely.  Good catch.  A further note is we may be able to do some more cleanup in the API ticket regarding this method/similar.

src/provider/tabbing/TabService.ts
```diff
@@ -206,114 +227,52 @@ export class TabService {
...
+        if (!activeDesktopWindow) return;
+
+        if (tabGroupUnderPoint) {
             // If we are over a tab group
 
             if (existingTabGroup !== tabGroupUnderPoint && tabAllowed) {
                 // And that tab group is not the one we are ejecting from
 
-                if (existingTabGroup) await existingTabGroup.removeTab(window);
+                if (existingTabGroup) await existingTabGroup.removeTab(activeDesktopWindow);
```
pjbroadbent: Just as a code style/consistency thing, can you make these `if`s two-liner's, with braces?

src/provider/tabbing/TabService.ts
```diff
@@ -21,6 +21,28 @@ export interface TabTarget extends TargetBase {
      * Represents the target window tabbing space;
      */
     dropArea: Rectangle;
+
+    /**
+     * The group that has been selected as the target candidate.
+     *
+     * This is not the group that the user is currently dragging, it is the group that has been selected as the target.
+     */
+    group: DesktopSnapGroup;
```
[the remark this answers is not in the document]
ryanpetersonOF: Correct, removed.

src/provider/View.ts
```diff
@@ -1,99 +1,123 @@
...
-     * This will be called every time the active group gets moved/resized, and again when the transformation ends (with
-     * null for both args).
-     *
-     * This ensures that the active and target groups have the correct opacity effects applied, and updates the snap
-     * preview.
-     *
-     * SnapView also stores these parameters as members. This allows it to revert the active/target windows to their
-     * original opacities once the active/target group(s) change or get reset.
-     */
     public update(activeGroup: DesktopSnapGroup|null, target: Target|null): void {
```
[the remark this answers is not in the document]
ryanpetersonOF: We dont store the active group (only activeWindow) in the targets.  But we can still remove this param and swap it with `target.activeWindow.snapGroup` in the method which I have done.

src/provider/View.ts
```diff
@@ -1,99 +1,123 @@
...
-     * @param group The activeGroup being dragged by the user.
-     * @param applyOnTop Apply alwaysOnTop or not.
-     */
-    private setAlwaysOnTop(group: DesktopSnapGroup|null, applyOnTop: boolean): void {
-        if (group) {
-            if (applyOnTop) {
-                group.windows.forEach((window: DesktopEntity) => {
-                    window.applyOverride('alwaysOnTop', true);
+    private setOpacity(entity: DesktopItem|null, transparent: boolean): void {
+        if (entity) {
```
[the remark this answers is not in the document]
ryanpetersonOF: We could null check outside, except we call it almost 8 times (including `setAlwaysOnTop` which has the same null check) passing in the previous / current activeItem/targetItem.  These can be null based on different factors.  It would pretty much require check the null each time it's called, hence it is moved into the method itself.

src/provider/APIHandler.ts
```diff
@@ -302,7 +302,8 @@ export class APIHandler {
         const target = tabService.getTarget(tab);
         tabService.dragWindowManager.hideWindow();
 
-        if (!target) return;
-        await tabService.applyTabTarget(target);
+        if (target) {
+            await tabService.applyTabTarget(target);
```
tomrobertsOF: Would prefer this to be `return` rather than `await` since it's at the end of the block, but will let it slide this time.

src/provider/model/MouseTracker.ts
pjbroadbent: Can you re-check this?

PR #223, opened 2019-01-17
[SERVICE-301] Enforce group validation on SnapGroups after workspace restore

tomrobertsOF:
Made some tweaks/improvements/fixes to the validation mechanism. 
`restoreGroup` now triggers a validate on all groups before returning.


src/provider/model/DesktopSnapGroup.ts
```diff
@@ -151,6 +167,79 @@ export class DesktopSnapGroup {
         }
     }
 
+    public validate(): void {
+        this._validateGroup.call();
+    }
+
+    private validateGroupInternal(): void {
+        // Ensure 'group' is still a valid, contiguous group.
+        // NOTE: 'modifiedWindow' may no longer exist (if validation is being performed because a window was closed)
```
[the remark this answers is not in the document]
tomrobertsOF: Done

src/provider/snapanddock/SnapService.ts
```diff
@@ -191,31 +181,11 @@ export class SnapService {
...
 
     private onGroupModified(group: DesktopSnapGroup, modifiedWindow: DesktopWindow): void {
-        if (group.windows.includes(modifiedWindow)) {
-            // If a validate is already scheduled, postpone it. But no need to trigger a validation.
-            this._validateGroups.postpone();
-        } else {
-            // Window has been removed from group, definitely need to validate.
-            this._validateGroups.call(group);
-        }
+        // Validation is handled at the group level. Need to think if anything else should happen here.
```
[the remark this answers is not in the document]
tomrobertsOF: done

src/provider/model/DesktopSnapGroup.ts
```diff
@@ -1,12 +1,22 @@
...
 import {DesktopEntity} from './DesktopEntity';
 import {DesktopTabGroup} from './DesktopTabGroup';
-import {DesktopWindow, EntityState, eTransformType, Mask, WindowIdentity, WindowMessages} from './DesktopWindow';
+import {DesktopWindow, EntityState, eTransformType, Mask, WindowMessages} from './DesktopWindow';
 
 export class DesktopSnapGroup {
+    /**
+     * Any windows less than this distance apart will be considered as touching for the purposes of the validateGroup
+     *
+     * This is a workaround for runtime issues due to be fixed in v37.
```
MichaelMCoates: What runtime issue is this, by the way? One case of this will be covered by SERV-302

tomrobertsOF: Not sure exactly, this was just copied with the rest of the code from SnapService. @pjbroadbent may be the best to answer.

tomrobertsOF:
Conflicts resolved

src/provider/model/DesktopSnapGroup.ts
```diff
@@ -151,6 +167,78 @@ export class DesktopSnapGroup {
...
+
+        while (unvisited.length > 0) {
+            const visited: DesktopEntity[] = [];
+            dfs(unvisited[0], visited);
+            contiguousSets.push(visited);
+        }
+
+        return contiguousSets;
+
+        function dfs(startWindow: DesktopEntity, visited: DesktopEntity[]) {
```
ryanpetersonOF: Can you give this a better name?

src/provider/model/DesktopSnapGroup.ts
pjbroadbent: The issue was that when resizing docked groups, the edges wouldn't always exactly align. Windows could be docked, but their edges would be off by up to ~7 pixels ish. The offset was to allow for 7px on each window.

That issue should now be fixed and this can likely be removed. @tomrobertsOF can you remove and do a quick check? Any issues whatsoever, and can defer the clean-up to some other story.

tomrobertsOF: Gave it a quick test and seems fine.

tomrobertsOF: From memory, the issue was particularly bad because of the problems with resizing framed window groups on win10.

src/provider/model/DesktopSnapGroup.ts
tomrobertsOF: Done

src/provider/workspaces/restore.ts
```diff
@@ -119,7 +119,7 @@ export const restoreLayout = async(payload: Layout, identity: Identity): Promise
             const mainWindowModel = model.getWindow(app.mainWindow);
             await tabService.removeTab(app.mainWindow);
             if (mainWindowModel!.snapGroup.length > 1) {
```
ryanpetersonOF: I know you didn't write this line, but since you're here:  would it make more sense to check `mainWindowModel` in the if?  If in the very small chance something happens to the window after `isRunning` check and getWindow returns `null`, we're going to get errors here.

```
if(mainWindowModel && mainWindowModel.snapGroup.length > 1) {
    await mainWindowModel.setSnapGroup(new DesktopSnapGroup());
}
```

Then we could remove the `!`s.  Same with `childWindowModel` in `src/provider/workspaces/utils.ts` right below.

tomrobertsOF: Good catch buddy! Done!

PR #224, opened 2019-01-21
[SERVICE-306] Config-based window register/de-register

pjbroadbent:
The config store is now "source of truth" for window register/de-register state. `DesktopModel` will use the config store to determine if it should register a window.

Model also uses a watch listener to detect when applications add/remove rules affecting the `enabled` property, and will then register/de-register any affected windows.

Have tested registering/de-registering windows long after their creation time, including de-registering windows that are snapped and/or tabbed.

Changes also include some update to signals - clarified behaviour of aggregators, added some common/re-usable aggregator functions.


src/provider/model/DesktopModel.ts
```diff
@@ -39,32 +53,33 @@ export class DesktopModel {
 
         const serviceUUID: string = fin.Application.me.uuid;
 
+        // Set built-in rules for determining if a window should be registered
+        const errorWindowSpec: RegEx = {expression: 'error-app-[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}', flags: 'g'};
+        config.addRule({level: 'service'}, {level: 'application', uuid: serviceUUID}, {enabled: false});
+        config.addRule({level: 'service'}, {level: 'window', uuid: serviceUUID, name: {expression: 'Placeholder-.*'}}, {enabled: true});
+        config.addRule({level: 'service'}, {level: 'application', uuid: errorWindowSpec}, {enabled: false});
```
pjbroadbent: I replaced the existing hand-coded rules for service windows with rules on the config store. These rules will need to be evaluated for every `query`/`queryPartial` call on the store.

Feedback welcome on whether this is "streamlining" or "over-engineering".

test/provider/deregisterConfig.test.ts
```diff
@@ -0,0 +1,100 @@
...
+        this.config.removeFromSource({level: 'window', uuid: 'testApp', name: 'testWindow2'});
+    });
+    await Promise.all(t.context.windows.map(win => win.close()));
+});
+
+async function addRuleToProvider(scope: Scope, config: ConfigurationObject): Promise<void> {
+    return executeJavascriptOnService(function(this: ProviderWindow, data) {
+        this.config.add(data.scope, data.config);
+    }, {scope, config});
+}
```
ryanpetersonOF: Is this something that can be used in other tests in the future?  If so might be worth putting into the `utils`

tomrobertsOF:
Looks great! Much cleaner way of doing it.

src/provider/model/DesktopModel.ts
```diff
@@ -197,34 +216,89 @@ export class DesktopModel {
...
-
-    private addWindow(window: Window): Promise<DesktopWindow|null> {
-        // Set the window as pending registration  (Fix for race condition between register/deregister)
-        const identity: WindowIdentity = window.identity as WindowIdentity;
-        this._pendingRegistrations.push({...identity});
+        const window: Window = fin.Window.wrapSync(identity);
         return DesktopWindow.getWindowState(window).then<DesktopWindow|null>((state: EntityState): DesktopWindow|null => {
-            if (!this._pendingRegistrations.some(w => w.name === identity.name && w.uuid === identity.uuid)) {
+            if (!this._config.queryPartial({level: 'window', ...identity}, enabledMask).enabled) {
                 // If pendingRegistrations does not contain the window, then deregister has been called on it
```
MichaelMCoates: Leftover comment.

src/provider/workspaces/create.ts
```diff
@@ -55,8 +49,8 @@ export const getCurrentLayout = async(): Promise<Layout> => {
 
         tabGroup.tabs.forEach((tabWindow: WindowIdentity) => {
             // Filter tabs out if either the window itself or its parent is deregistered from Save and Restore
-            const parentIsDeregistered = inWindowObject({uuid: tabWindow.uuid, name: tabWindow.uuid}, deregisteredWindows);
-            const windowIsDeregistered = inWindowObject(tabWindow, deregisteredWindows);
+            const parentIsDeregistered = !!model.getWindow({uuid: tabWindow.uuid, name: tabWindow.uuid});
+            const windowIsDeregistered = !!model.getWindow(tabWindow);  // Should always be true..?
```
MichaelMCoates: Correct, this can be removed.

src/provider/workspaces/create.ts
```diff
@@ -55,8 +49,8 @@ export const getCurrentLayout = async(): Promise<Layout> => {
 
         tabGroup.tabs.forEach((tabWindow: WindowIdentity) => {
             // Filter tabs out if either the window itself or its parent is deregistered from Save and Restore
-            const parentIsDeregistered = inWindowObject({uuid: tabWindow.uuid, name: tabWindow.uuid}, deregisteredWindows);
-            const windowIsDeregistered = inWindowObject(tabWindow, deregisteredWindows);
+            const parentIsDeregistered = !!model.getWindow({uuid: tabWindow.uuid, name: tabWindow.uuid});
```
MichaelMCoates: Shouldn't these be `!!!`?

If `model.getWindow` returns a window, !!DesktopWindow === true, which means the parent isn't deregistered, but `parentIsDeregistered` is set to true.

test/provider/deregisterConfig.test.ts
pjbroadbent: Probably. Will figure out on the next story - not sure if we'll end up with enough helpers for a "config utils", or if we'll need to just find somewhere within the existing structure for this.

PR #225, opened 2019-01-21
[SERVICE-299] Publish a version of the client that can be imported directly from the browser

ryanpetersonOF:
**Do Not Merge Yet**  Pending API changes

Creates a minified client library which can be included via `script` tag.  The location will be on the CDN as `https://cdn.openfin.co/services/openfin/layouts/<VERSION>/openfin-layouts.js`.  Once included, the library will be accessible via global var `OpenFinLayouts`.  This can be tested in the demo app using 'Client Lib Included' button.

### DIFF
--- a/res/demo/libScriptIncluded.html
+++ b/res/demo/libScriptIncluded.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Openfin Layouts Included</title>
+    <script src="http://localhost:1337/client/openfin-layouts.js"></script>
+</head>
+<body>
+    <script>
+        const randomColor = () => {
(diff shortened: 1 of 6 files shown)


res/demo/libScriptIncluded.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>Openfin Layouts Included</title>
+    <script src="http://localhost:1337/client/openfin-layouts.js"></script>
```
tomrobertsOF: Does this work when running the demo targeting a cdn version? (e.g. `npm start --version staging`)

ryanpetersonOF: Yes, the express server is still started so all is well.

PR #229, opened 2019-01-25
[SERVICE-313] Added "test teardown" util

pjbroadbent:
Added an 'afterEach' hook to run after each integration test. The hook will ensure that the desktop state has been completely reset before the next test starts.

If there is any left-over state (such as a window being open that shouldn't be), then the util will repair the issue to get the test runner back into it's default state, and then fail the previous test with a description of the problem.

There is one exception to the "fail" rule - due to a known issue where closing multiple tabs simultaneously may not destroy the tabstrip window, any remaining tabstrip windows will not cause the previous test to fail. The util will still clean-up the left-over tabstrip window(s). The root cause of this issue will be investigated in a future story.

### DIFF
--- a/res/test/provider.html
+++ b/res/test/provider.html
@@ -10,7 +10,7 @@
             channelProvider.register('execute-javascript', async (payload) => {
                 try {
                     result = await new Function("return " + payload.script)()(payload.data)
-                    return {success: true, result: JSON.stringify(result)};
+                    return {success: true, result};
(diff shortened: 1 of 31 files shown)


pjbroadbent:
Merging with just the one approval as the current test instability is blocking feature work.

@tomrobertsOF @ryanpetersonOF Any feedback, post here and will address in a future PR.

PR #233, opened 2019-01-28
[SERVICE-302] Properly restore windows that were formerly a part of a tab group.

MichaelMCoates:
- Previously, if a window was a part of a tab group that was going to be disbanded (say, if it is tabbed to the child window of a deregistered App), that window would restore at a smaller size. 
- Now, windows restore at their proper size.

### DIFF
--- a/src/provider/workspaces/create.ts
+++ b/src/provider/workspaces/create.ts
@@ -10,7 +10,7 @@ import {WindowIdentity} from '../model/DesktopWindow';
 import {promiseMap} from '../snapanddock/utils/async';
 
 import {getGroup} from './group';
-import {addToWindowObject, inWindowObject, parseVersionString, wasCreatedFromManifest, wasCreatedProgrammatically, WindowObject, adjustSizeOfFormerlyTabbedWindows} from './utils';
+import {addToWindowObject, adjustSizeOfFormerlyTabbedWindows, inWindowObject, parseVersionString, wasCreatedFromManifest, wasCreatedProgrammatically, WindowObject} from './utils';
 
 // This value should be updated any time changes are made to the layout schema.
 // Major version indicates breaking changes.
@@ -116,13 +116,13 @@ export const getCurrentLayout = async(): Promise<Layout> => {
             const mainWindow: LayoutWindow = {...windowInfo.mainWindow, ...mainWindowLayoutData};
             const mainWinIdentity = {uuid, name: mainWindow.name};
             adjustSizeOfFormerlyTabbedWindows(mainWinIdentity, formerlyTabbedWindows, mainWindow);
-            
+
             // Filter for deregistered child windows
             windowInfo.childWindows = windowInfo.childWindows.filter((win: WindowDetail) => {
                 const isDeregistered = inWindowObject({uuid, name: win.name}, deregisteredWindows);
                 return isDeregistered ? false : true;
             });
-            
+
             // Grab the layout information for the child windows
             const childWindows: LayoutWindow[] = await promiseMap(windowInfo.childWindows, async (childWin: WindowDetail) => {
                 const {name} = childWin;
@@ -211,6 +211,6 @@ const getLayoutWindowData = async(ofWin: Window, tabbedWindows: WindowObject): P
 
     // If a window is tabbed (based on filtered tabGroups), tab it.
     const isTabbed = inWindowObject(ofWin.identity, tabbedWindows) ? true : false;
-    
+
     return {info, uuid, windowGroup: filteredWindowGroup, frame: applicationState.frame, state: options.state, isTabbed, isShowing: !applicationState.hidden};
 };
--- a/src/provider/workspaces/create.ts
+++ b/src/provider/workspaces/create.ts
@@ -120,14 +120,14 @@ export const getCurrentLayout = async(): Promise<Layout> => {
             // Filter for deregistered child windows
             windowInfo.childWindows = windowInfo.childWindows.filter((win: WindowDetail) => {
                 const isDeregistered = inWindowObject({uuid, name: win.name}, deregisteredWindows);
-                return isDeregistered ? false : true;
+                return !isDeregistered;
             });
 
             // Grab the layout information for the child windows
             const childWindows: LayoutWindow[] = await promiseMap(windowInfo.childWindows, async (childWin: WindowDetail) => {
                 const {name} = childWin;
                 const childWinIdentity = {uuid, name};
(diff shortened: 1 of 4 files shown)


src/provider/workspaces/create.ts
```diff
@@ -111,23 +114,24 @@ export const getCurrentLayout = async(): Promise<Layout> => {
...
+            adjustSizeOfFormerlyTabbedWindows(mainWinIdentity, formerlyTabbedWindows, mainWindow);
 
             // Filter for deregistered child windows
             windowInfo.childWindows = windowInfo.childWindows.filter((win: WindowDetail) => {
                 const isDeregistered = inWindowObject({uuid, name: win.name}, deregisteredWindows);
-                if (isDeregistered) {
-                    return false;
-                }
-                return true;
+                return isDeregistered ? false : true;
```
ryanpetersonOF: Can do `return !isDeregistered` ?

src/provider/workspaces/create.ts
```diff
@@ -111,23 +114,24 @@ export const getCurrentLayout = async(): Promise<Layout> => {
...
             });
 
             // Grab the layout information for the child windows
-            const childWindows: LayoutWindow[] = await promiseMap(windowInfo.childWindows, async (win: WindowDetail) => {
-                const {name} = win;
-                const ofWin = await fin.Window.wrap({uuid, name});
+            const childWindows: LayoutWindow[] = await promiseMap(windowInfo.childWindows, async (childWin: WindowDetail) => {
+                const {name} = childWin;
+                const childWinIdentity = {uuid, name};
+                const ofWin = await fin.Window.wrap(childWinIdentity);
```
ryanpetersonOF: A minor one but `await fin.Window.wrap` is the same as `fin.Window.wrapSync()` but without the promise.  It might be "cleaner" to use here.

src/provider/workspaces/utils.ts
```diff
@@ -234,4 +237,22 @@ export function parseVersionString(versionString: string): SemVer {
     }
 
     return {major: Number.parseInt(match[1], 10), minor: Number.parseInt(match[2], 10), patch: Number.parseInt(match[3], 10)};
+}
+
+export function adjustSizeOfFormerlyTabbedWindows(winIdentity: WindowIdentity, formerlyTabbedWindows: WindowObject, layoutWindow: LayoutWindow|WindowDetail) {
+    if (inWindowObject(winIdentity, formerlyTabbedWindows)) {
+        const tabWindow = model.getWindow(winIdentity);
```
ryanpetersonOF: Wrapping everything after this line in `if(tabWindow) {...}` will remove the need to put `!` after everything.  Same for `tabGroup`

MichaelMCoates: got it!

src/provider/workspaces/create.ts
```diff
@@ -111,23 +114,24 @@ export const getCurrentLayout = async(): Promise<Layout> => {
             const mainOfWin: Window = await ofApp.getWindow();
             const mainWindowLayoutData = await getLayoutWindowData(mainOfWin, tabbedWindows);
             const mainWindow: LayoutWindow = {...windowInfo.mainWindow, ...mainWindowLayoutData};
+            const mainWinIdentity = {uuid, name: mainWindow.name};
+            adjustSizeOfFormerlyTabbedWindows(mainWinIdentity, formerlyTabbedWindows, mainWindow);
```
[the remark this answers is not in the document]
MichaelMCoates: Considering it's used twice, and the `getCurrentLayout` function is pretty fat, I thought it was cleaner to do it like this.

src/provider/workspaces/create.ts
```diff
@@ -111,23 +114,24 @@ export const getCurrentLayout = async(): Promise<Layout> => {
...
 
             // Filter for deregistered child windows
             windowInfo.childWindows = windowInfo.childWindows.filter((win: WindowDetail) => {
                 const isDeregistered = inWindowObject({uuid, name: win.name}, deregisteredWindows);
-                if (isDeregistered) {
-                    return false;
-                }
-                return true;
+                return !isDeregistered;
```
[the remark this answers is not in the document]
MichaelMCoates: Oh nice, changing.

test/demo/workspaces/formerTabGroupRestoreResize.test.ts
```diff
@@ -0,0 +1,59 @@
...
+formerTabGroupTestOptionsArray.push({apps: combinedProgrammaticApps, tabWindowGrouping: windowGrouping});
+formerTabGroupTestOptionsArray.push({apps: combinedManifestApps, tabWindowGrouping: windowGrouping});
+
+
+testParameterized<CreateAppData, AppContext>(
+    (testOptions: CreateAppData): string =>
+        `Tab SaveAndRestore - ${testOptions.apps[0].createType === 'manifest' ? 'Manifest' : 'Programmatic'} - Formerly tabbed window resize`,
+    formerTabGroupTestOptionsArray,
+    createAppTest(async (t, applicationData: CreateAppData) => {
+        const group = applicationData.tabWindowGrouping![0];
```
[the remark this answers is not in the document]
MichaelMCoates: gotcha!

test/demo/workspaces/formerTabGroupRestoreResize.test.ts
```diff
@@ -0,0 +1,59 @@
...
+const windowGrouping = [[0, 2]];
+
+formerTabGroupTestOptionsArray.push({apps: combinedProgrammaticApps, tabWindowGrouping: windowGrouping});
+formerTabGroupTestOptionsArray.push({apps: combinedManifestApps, tabWindowGrouping: windowGrouping});
+
+
+testParameterized<CreateAppData, AppContext>(
+    (testOptions: CreateAppData): string =>
+        `Tab SaveAndRestore - ${testOptions.apps[0].createType === 'manifest' ? 'Manifest' : 'Programmatic'} - Formerly tabbed window resize`,
+    formerTabGroupTestOptionsArray,
```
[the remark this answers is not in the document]
MichaelMCoates: Yeah I was having trouble with this too. I've changed it to expanding out the individual test params, but I don't think that solves the problem. I'd probably have to change up my app creation/set-up code to be more semantic.

src/provider/workspaces/create.ts
MichaelMCoates: Got it, thanks!

src/provider/workspaces/create.ts
MichaelMCoates: Great!

PR #240, opened 2019-02-03
[SERVICE-291] Load config from Application Manifests

pjbroadbent:
Added a "Loader" component, separate from the main config store, that will listen to application lifecycle events and then add and remove config to the store as necessary.

The loader will fetch the manifest of each application that is created. If the application manifest doesn't reference the service, or references the service but doesn't provide any configuration data, then the loader will ignore it. If an application does define config for the service, it will add that config to the store and then listen for when the application closes.

When the application closes, the loader will instruct the config to delete any configuration data that was defined by that application. There is one exception to this - since applications that were launched programmatically can't define their own config (as there is no manifest), the loader will only unload an application's config once that application has exited, *and* the application has no running child applications ("child" being defined by `getInfo().parentUuid`).

To facilitate ongoing testing of this mechanism, the main Layouts Manager window in the demo app now de-registers itself via it's manifest, rather than the `deregister` client API.


tomrobertsOF:
Looks great! Can test further once it's in the integration branch.

src/provider/config/ConfigUtil.ts
```diff
@@ -72,6 +72,31 @@ type MaskedHelper<T, M, K extends(keyof T)&(keyof M)> = {
...
  */
 export class ConfigUtil {
+    /**
+     * Takes a scope or rule and converts it to a canonical string that can be used as a cache key, or in logging
+     * output.
+     *
+     * @param scope The scope/rule to stringify
+     */
+    public static getId(scope: Rule|Scope): string {
+        function stringifyParam(param: string|RegEx): string {
```
MichaelMCoates: Would this redefine the `stringifyParam` function every time the `getId` function is called? Would it be better to define it outside the function?

src/provider/config/Loader.ts
```diff
@@ -0,0 +1,195 @@
...
+                parent = parent.parent;
+                console.log('Clean up parent', parent.scope.uuid);
+                this.cleanUpApplicationConfig(parent);
+            }
+        }
+    }
+
+    /**
+     * Cleans-up any config within the store for applications that are no longer running and have no child applications
+     * that may be dependant on rules defined in a parent application.
```
MichaelMCoates: `dependent` is the adjective in both British and American English.

src/provider/config/ConfigUtil.ts
pjbroadbent: As it's a function declaration, rather than a `const stringifyParam = function() { ...`, I _guess_ it's only actually defined at parse-time?

But no idea. Will move it out to a private function to be safe.

PR #242, opened 2019-02-04
[SERVICE-308] Validate desktop after monitor resolution change

tomrobertsOF:
On `monitor-info-change` event we now globally validate all tab- and snapGroups.

SnapGroups are validated with the same logic as in other places: group is split into any remaining contiguous sets of windows.

For tabGroups, monitor resizes generally result in all state being correct with the exception of the tabstrip. When validating we check the relative position of the tabStrip to the group and move it back into position.

Note Re: testing - I discussed with Phil and we think that it would be unwise to automate screen resizing on the test slaves with the current setup. For the time being this will need to be tested manually.

### DIFF
--- a/src/provider/model/DesktopModel.ts
+++ b/src/provider/model/DesktopModel.ts
@@ -67,6 +67,15 @@ export class DesktopModel {
                 }
             });
         });
+
+        // Validate everything on monitor change, as groups may become disjointed
+        fin.System.addListener('monitor-info-changed', async () => {
+            // Validate all tabgroups
+            this.tabGroups.map(g => g.validate());
+
+            // Validate all snap groups
+            this.snapGroups.map(g => g.validate());
+        });
     }
 
     public get mouseTracker(): MouseTracker {
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -3,7 +3,8 @@ import {_Window} from 'hadouken-js-adapter/out/types/src/api/window/window';
 import {JoinTabGroupPayload, TabGroupEventPayload, TabPropertiesUpdatedPayload} from '../../client/tabbing';
 import {ApplicationUIConfig, TabProperties, WindowIdentity} from '../../client/types';
 import {Signal1} from '../Signal';
-import {Point} from '../snapanddock/utils/PointUtils';
+import {Debounced} from '../snapanddock/utils/Debounced';
+import {Point, PointUtils} from '../snapanddock/utils/PointUtils';
 import {Rectangle, RectUtils} from '../snapanddock/utils/RectUtils';
 
 import {DesktopEntity} from './DesktopEntity';
@@ -55,6 +56,7 @@ export class DesktopTabGroup implements DesktopEntity {
 
     private _config: ApplicationUIConfig;
 
+    private _validateGroup: Debounced<() => void, DesktopTabGroup, []>;
 
     /**
      * Constructor for the TabGroup Class.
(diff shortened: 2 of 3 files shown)


src/provider/model/DesktopWindow.ts
```diff
@@ -906,11 +938,23 @@ export class DesktopWindow implements DesktopEntity {
...
+            DesktopWindow.activeTransactions.filter((transaction: Transaction) => transaction.windows.some(w => !!targetWindow && w._id === targetWindow._id));
+        if (relevantTransactions.length > 0) {
+            console.log('Window currently in a transaction. Ignoring window group changed event: ', event);
+            relevantTransactions.forEach(t => {
+                t.remove.postpone();
+            });
+            return;
+        }
+
+        console.log('Received window group changed event: ', event);
```
ryanpetersonOF: Is this logging absolutely required?  Same with above.

tomrobertsOF: Yes

src/provider/model/DesktopWindow.ts
```diff
@@ -170,12 +184,30 @@ export class DesktopWindow implements DesktopEntity {
...
+        // Create a transaction object and add it to the active transactions list.
+        // The 'remove' property looks a bit strange but effectively lets the object remove itself
+        // from the list when prompted.
+        const transaction: Transaction = {
+            windows,
+            remove: new Debounced(
+                () => {
+                    const indexToRemove = this.activeTransactions.indexOf(transaction);
+                    if (indexToRemove >= 0) {
+                        this.activeTransactions.splice(indexToRemove, 1);
```
MichaelMCoates: Is there any way that the splice could end up splicing at the wrong index? I could see multiple transactions being scheduled simultaneously, but I do see that this is only really being used for the `tabStripOffset`

tomrobertsOF: The index selection and slicing is fully synchronous. There is no way for any other code to jump in the middle of those two lines (or if it did, it would be a pretty major engine bug).

src/provider/model/DesktopTabGroup.ts
```diff
@@ -588,4 +596,19 @@ export class DesktopTabGroup implements DesktopEntity {
...
+    // moves them so that they are
+    private async validateGroupInternal() {
+        const tabStripOffset: Point<number> = PointUtils.difference(
+            this._window.currentState.center,
+            {x: this.currentState.center.x, y: this.currentState.center.y - this.currentState.halfSize.y + this.config.height / 2});
+
+        if (PointUtils.lengthSquared(tabStripOffset) > 0) {
+            console.log('TabGroup disjointed. Moving tabstrip back to group.', this.id);
+            await DesktopWindow.transaction([this._window], async (wins: DesktopWindow[]) => {
+                await wins[0].applyOffset(tabStripOffset);
```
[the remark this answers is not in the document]
tomrobertsOF: Not that I've seen in testing. It seems to preserve size and position the top-left proportional to where it was on the previous resolution.

src/provider/model/DesktopModel.ts
```diff
@@ -67,6 +67,15 @@ export class DesktopModel {
...
             });
         });
+
+        // Validate everything on monitor change, as groups may become disjointed
+        fin.System.addListener('monitor-info-changed', async () => {
+            // Validate all tabgroups
+            this.tabGroups.map(g => g.validate());
+
+            // Validate all snap groups
+            this.snapGroups.map(g => g.validate());
```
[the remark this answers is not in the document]
tomrobertsOF: Both of these basically just `call()` a debounce, which has no way currently to "await" the execution of the callback. 

As for the danger of doing both "in parallel", there shouldn't be much of an issue. In tab validate we assume the current state of the group to be true and move the tabstrip back to that position. If a group validate happens in the meantime, the state of the group as a DesktopEntity is still correct, so should be validated properly.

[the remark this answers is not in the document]
tomrobertsOF: That comment came out as a bit of a jumbled mess. Let me know if you want me to explain face-to-face.

PR #243, opened 2019-02-04
[SERVICE-317] Tab tearout will only work on one monitor at a time.

ryanpetersonOF:
Allows for the drag window to cover any monitor enabling multimonitor tabbing.

### DIFF
--- a/src/provider/tabbing/DragWindowManager.ts
+++ b/src/provider/tabbing/DragWindowManager.ts
@@ -6,6 +6,7 @@ import {WindowIdentity} from '../../client/types';
 import {DesktopModel} from '../model/DesktopModel';
 import {DesktopWindow} from '../model/DesktopWindow';
 import {Signal0, Signal2} from '../Signal';
+import { MonitorInfo, DipRect } from 'hadouken-js-adapter/out/types/src/api/system/monitor';
 
 /**
  * Handles the Drag Window which appears when API drag and drop is initialized.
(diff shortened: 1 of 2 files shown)


src/provider/tabbing/DragWindowManager.ts
```diff
@@ -120,8 +139,22 @@ export class DragWindowManager {
...
+     */
+    private async setWindowBounds() {
+        const monitorInfo: MonitorInfo = await fin.System.getMonitorInfo();
+        this._virtualScreen = monitorInfo.virtualScreen;
+
+        this._window.setBounds(
+            this._virtualScreen.left,
+            this._virtualScreen.top,
+            Math.abs(this._virtualScreen.left - this._virtualScreen.right),
+            Math.abs(this._virtualScreen.top - this._virtualScreen.bottom));
```
pjbroadbent: Could simplify a bit - right/bottom will always be greater than left/top.

ryanpetersonOF: Fixed 😅

tomrobertsOF:
Do we have any way of running those tests on multiple monitors in CI? Either way, look like a good solution.

src/provider/tabbing/DragWindowManager.ts
```diff
@@ -29,16 +30,32 @@ export class DragWindowManager {
...
     private _model: DesktopModel;
 
     constructor(model: DesktopModel) {
         this._model = model;
         this._sourceWindow = null;
         this._hideTimeout = -1;
         this.createDragWindow();
+
+        fin.System.addListener('monitor-info-changed', () => {
+            this.setWindowBounds();
```
tomrobertsOF: Does the event object have the virtualScreen info, and if so could we use that to avoid an extra async call into the core?

tomrobertsOF: Nice. Much better after the change

ryanpetersonOF: Fixed

PR #244, opened 2019-02-04
[SERVICE-305] Replace existing URL-based parameters with JSON config

pjbroadbent:
Removed the previous (temporary) service-wide flags with the config mechanism.

The services that had toggle-able features now apply these feature flags on a per-window basis, so these features can be disabled for specific windows/apps without affecting any unrelated windows.

### DIFF
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -11,6 +11,7 @@ import {DesktopModel} from './DesktopModel';
 import {DesktopSnapGroup} from './DesktopSnapGroup';
 import {DesktopTabstripFactory} from './DesktopTabstripFactory';
 import {DesktopWindow, EntityState, eTransformType, Mask, WindowMessages} from './DesktopWindow';
+import { Scope } from '../../../gen/provider/config/scope';
 
 /**
  * Handles functionality for the TabSet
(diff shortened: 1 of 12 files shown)


src/provider/model/DesktopTabGroup.ts
```diff
@@ -141,6 +142,19 @@ export class DesktopTabGroup implements DesktopEntity {
         return this._tabs;
     }
 
+    /**
+     * Scope isn't really strictly defined for a tab group...
+     *
+     * Will assume that the tab group should follow the config rules for the active tab, in reality it should be some
```
tomrobertsOF: What happens if you have two tabs, one with snapping enabled and one without; with the snappable tab active, snap the tab group to some other window, then with to the non-snappable tab?

pjbroadbent: The tab group would only snap when the `{snap: true}` tab is active; when the `{snap: false}` tab is active, the tabgroup will act as if snapping were disabled for the whole tab group.

Thinking about docking... The only times we check the config are as part of the `getTarget`/`applyTarget` system. If the docking-enabled check passes at that point, then those windows will dock as usual. Switching tabs won't cause the group to detach, neither will any kind of group validation action - once SnapService decides that two windows should dock, they'll remain docked as per the existing rules.

pjbroadbent:
Since the config feature adds a good chunk more "state" to the provider, have added the store and loader to the list of checks that get performed during teardown.

Haven't yet added anything to restore the state if it gets corrupted (since repairing the default set of rules is fiddly without duplicating them), but will look into doing that if we start seeing these warnings within test runner output.

src/provider/model/DesktopTabGroup.ts
MichaelMCoates: Just tested this behavior and it works as intended.

MichaelMCoates:
Build failing. One clear example is `{"dock": false}` was set earlier in the tests, but spills over to the `SnapAndDock` tests. Looks like windows are getting cleaned up before they can remove their rules from the store.

pjbroadbent:
Oh, oops....

Since the 'loader' story, config will get cleaned up when a window exits - but then in the dockingDisabled tests, I attached those rules to the testApp window, which stays open for the whole test run.

Have moved the rules to the test windows, renamed those windows in case of future "config leaking" issues, and the config teardown will now remove any rules from a non-`service` source scope.

Currently re-running on CI.

PR #248, opened 2019-02-05
[SERVICE-312] Tabstrip can become detached when calling `undockGroup` with certain configurations of snapped tabGroups

tomrobertsOF:
Quick fix for this issue. We now leave one window in the previous dock group when exploding.


tomrobertsOF:
Tests are passing on CI. Will merge if there are no objections.

PR #251, opened 2019-02-06
[SERVICE-321] Quick mouse movements when tabbing can cause disjointed tabStrip

tomrobertsOF:
Now call `disableFrame` on the tabstrip on creation. This also works nicely with the disabledFrameGrouping in v38


tomrobertsOF:
Test have passed on CI. Will merge in a few minutes if no objections.

PR #267, opened 2019-02-08
[SERVICE-336] Clicking restore in the middle of another restore breaks restoration

bryangaleOF:
This still needs full testing, but would appreciate eyes on this.

There's four real parts to this

1. At the start and end of the restoreWorkspace method, mark and unmark that a restore is in progress, and if the restoreWorkspace  method is called again during a restore, throw an exception

2. Add a timeout when waiting for a  WorkspaceAPI.RESTORE_HANDLER call to the client app via sendToClient to return

3. Add a timeout when waiting for a client app to notify the provider that it is ready to restore

These 3 together /should/ be enough to ensure that we don't start a restore when we are already mid-restore but will also never block restores permanently due to a misbehaving client app. However, for caution, we also have:

4. A global timeout on the marker that restore is running, so that we definitely never permanently block restores

As part of this PR, I've also split up the restoreWorkspace to make it more manageable

### DIFF
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -60,7 +60,7 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
 
     await createAllPlaceholders(layout);
 
-    const apps = await promiseMap(layout.apps, async(app: WorkspaceApp): Promise<WorkspaceApp> => await restoreApp(app, startupApps));
+    const apps = await promiseMap(layout.apps, app => restoreApp(app, startupApps));
 
     // Wait for all apps to startup
     const startupResponses = await Promise.all(startupApps);
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -299,13 +299,13 @@ const restoreApp = async(app: WorkspaceApp, startupApps: Promise<WorkspaceApp>[]
 };
 
 const clientRestoreAppWithTimeout = async(app: WorkspaceApp, mayBeLegacyApp: boolean): Promise<WorkspaceApp> => {
-    const {uuid} = app;
+    const identity = {uuid: app.uuid, name: app.uuid};
 
     const defaultResponse = {...app, childWindows: []};
 
-    const sendToClientPromises = [apiHandler.sendToClient<WorkspaceApp, WorkspaceApp|false>({uuid, name: uuid}, WorkspaceAPI.RESTORE_HANDLER, app)];
+    const sendToClientPromises = [apiHandler.sendToClient<WorkspaceApp, WorkspaceApp|false>(identity, WorkspaceAPI.RESTORE_HANDLER, app)];
     if (mayBeLegacyApp) {
-        sendToClientPromises.push(apiHandler.sendToClient<WorkspaceApp, WorkspaceApp|false>({uuid, name: uuid}, LegacyAPI.RESTORE_HANDLER, app));
+        sendToClientPromises.push(apiHandler.sendToClient<WorkspaceApp, WorkspaceApp|false>(identity, LegacyAPI.RESTORE_HANDLER, app));
     }
 
     const responsePromise = Promise.race(sendToClientPromises).then((response: WorkspaceApp|false|undefined) => response ? response : defaultResponse);
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -13,8 +13,13 @@ import {SCHEMA_MAJOR_VERSION} from './create';
 import {regroupWorkspace} from './group';
 import {addToWindowObject, childWindowPlaceholderCheck, childWindowPlaceholderCheckRunningApp, createNormalPlaceholder, createTabbedPlaceholderAndRecord, inWindowObject, parseVersionString, positionWindow, SemVer, TabbedPlaceholders, wasCreatedProgrammatically, WindowObject} from './utils';
 
-const GLOBAL_RESTORE_TIMEOUT = 120000;
+// Duration in milliseconds that the entire Workspace restore may take, before we allow another restore to start
+const GLOBAL_EXCLUSIVITY_TIMEOUT = 120000;
+
+// Duration in milliseconds that we give a client app to startup when restoring a Workspace
 const CLIENT_STARTUP_TIMEOUT = 60000;
+
+// Duration in milliseconds that we give a client app to restore itself when restoring a Workspace
 const CLIENT_RESTORE_TIMEOUT = 60000;
 
 const appsCurrentlyStarting = new Map();
@@ -148,7 +153,7 @@ const startRestoreBlockerTimeout = (): void => {
                 appsToRestore.clear();
                 appsCurrentlyRestoring.clear();
             }
-        }, GLOBAL_RESTORE_TIMEOUT);
+        }, GLOBAL_EXCLUSIVITY_TIMEOUT);
     })();
 };
 
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -46,7 +46,7 @@ export const appReadyForRestore = async(uuid: string): Promise<void> => {
     }
 };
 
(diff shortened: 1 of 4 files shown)


src/provider/workspaces/restore.ts
```diff
@@ -13,50 +13,98 @@ import {SCHEMA_MAJOR_VERSION} from './create';
 import {regroupWorkspace} from './group';
 import {addToWindowObject, childWindowPlaceholderCheck, childWindowPlaceholderCheckRunningApp, createNormalPlaceholder, createTabbedPlaceholderAndRecord, inWindowObject, parseVersionString, positionWindow, SemVer, TabbedPlaceholders, wasCreatedProgrammatically, WindowObject} from './utils';
 
+const GLOBAL_RESTORE_TIMEOUT = 120000;
+const CLIENT_STARTUP_TIMEOUT = 60000;
+const CLIENT_RESTORE_TIMEOUT = 60000;
```
pjbroadbent: Can you explain what each of these is used for? We also have a `Constants.ts`, which may be a better place for these.

src/provider/workspaces/restore.ts
```diff
@@ -13,50 +13,98 @@ import {SCHEMA_MAJOR_VERSION} from './create';
...
+
+const appsCurrentlyStarting = new Map();
 const appsToRestore = new Map();
 const appsCurrentlyRestoring = new Map();
 
+let restoreBlocker: {}|null = null;
```
pjbroadbent: Can you explain the purpose of this.

Could we use the `Workspace` object as the token?

src/provider/workspaces/restore.ts
```diff
@@ -13,50 +13,98 @@ import {SCHEMA_MAJOR_VERSION} from './create';
...
+export const restoreWorkspace = async(payload: Workspace, identity: Identity): Promise<Workspace> => {
+    if (restoreBlocker !== null) {
+        throw new Error('Attempting to restore while restore in progress');
+    }
+
+    validatePayload(payload);
+
+    startRestoreBlockerTimeout();
+
+    const layout = payload;
```
pjbroadbent: Whilst we're changing this code, can we start using "workspace" in favour of "layout"?

src/provider/workspaces/restore.ts
```diff
@@ -13,50 +13,98 @@ import {SCHEMA_MAJOR_VERSION} from './create';
...
+    const startupApps: Promise<WorkspaceApp>[] = [];
+
+    await createAllPlaceholders(layout);
+
+    const apps = await promiseMap(layout.apps, async(app: WorkspaceApp): Promise<WorkspaceApp> => await restoreApp(app, startupApps));
```
pjbroadbent: The extra `async` and `await` here are a bit unnecessary, this line is equivalent to:
`const apps = await promiseMap(layout.apps, (app: WorkspaceApp): Promise<WorkspaceApp> => restoreApp(app, startupApps));`

For trivial lamba-type functions like this one, can probably omit the type declarations entirely?
`const apps = promiseMap(layout.apps, app => restoreApp(app, startupApps));`

src/provider/workspaces/restore.ts
bryangaleOF: Constants.ts is under snapanddock, so doesn't feel like the right place, but have commented these constants

src/provider/workspaces/restore.ts
bryangaleOF: Anticipating merge hell with SR-341 as is, so would prefer to leave this until after that's in

src/provider/workspaces/restore.ts
bryangaleOF: Have improved naming and commenting around this. Would prefer to leave implementation as is and not use Workspace token, as global knowledge is needed to reason about this being usable for this purpose

bryangaleOF:
Probably easier for PR-269/SERVICE-341 to go in before this (CC @MichaelMCoates)

bryangaleOF:
Jenkins now green for this

src/provider/workspaces/restore.ts
```diff
@@ -13,50 +13,104 @@ import {SCHEMA_MAJOR_VERSION} from './create';
...
+    return layout;
 };
 
-export const restoreApplication = async(layoutApp: WorkspaceApp, resolve: Function): Promise<void> => {
+const requestClientRestoreApp = async(layoutApp: WorkspaceApp, resolve: Function): Promise<void> => {
     const {uuid} = layoutApp;
     const name = uuid;
-    const defaultResponse: WorkspaceApp = {...layoutApp, childWindows: []};
     const appConnection = apiHandler.isClientConnection({uuid, name});
     if (appConnection) {
```
tomrobertsOF: Not part of your change, but here we assign to some consts and then only use them once in the following lines. Can this just be consolidated to something like
```
const {uuid} = layoutApp
if (apiHandler.isClientConnection({uuid: , name: uuid}) { ...
```

src/provider/workspaces/restore.ts
```diff
@@ -85,9 +139,26 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
...
+
+                appsCurrentlyStarting.clear();
+                appsToRestore.clear();
+                appsCurrentlyRestoring.clear();
+            }
+        }, GLOBAL_EXCLUSIVITY_TIMEOUT);
+    })();
+};
+
+const createAllPlaceholders = async(layout: Workspace): Promise<void> => {
```
tomrobertsOF: This function called `createAllPlaceholders` has an internally defined function also called `createAllPlaceholders`. May cause some confusion depending on scope. Maybe make the internal one `createPlaceholdersInternal` or something like that? Could possibly even just put it as an anonymous function inside the promiseMap.

src/provider/workspaces/restore.ts
```diff
@@ -158,89 +228,102 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
...
-                if (ofAppNotRunning) {
-                    await ofAppNotRunning.run().catch(console.log);
-                    const ofWindowNotRunning = await ofAppNotRunning.getWindow();
-                    await positionWindow(app.mainWindow);
-                }
-                // SHOULD WE RETURN DEFAULT RESPONSE HERE?!?
-                return defaultResponse;
+            if (ofAppNotRunning) {
+                await ofAppNotRunning.run().catch(console.log);
+                const ofWindowNotRunning = await ofAppNotRunning.getWindow();
```
tomrobertsOF: This is assigned but never used.

src/provider/workspaces/restore.ts
```diff
@@ -158,89 +228,102 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
...
-                    // If app created by manifest
-                    const {manifestUrl} = app;
-                    console.log('App has manifestUrl:', app);
-                    ofAppNotRunning = await fin.Application.createFromManifest(manifestUrl);
+                // Not connected to service
+                await positionWindow(app.mainWindow);
+                return defaultResponse;
+            }
+        } else {
+            let ofAppNotRunning: undefined|Application;
```
tomrobertsOF: More a matter of test, but can we make this `Application|undefined`

src/provider/workspaces/restore.ts
```diff
@@ -158,89 +228,102 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
...
+    } catch (e) {
+        console.error('Error restoring app', app, e);
+        return defaultResponse;
     }
+};
 
-    apiHandler.sendToAll('workspace-restored', layout);
+const clientRestoreAppWithTimeout = async(app: WorkspaceApp, mayBeLegacyApp: boolean): Promise<WorkspaceApp> => {
+    const {uuid} = app;
+    const name = uuid;
```
tomrobertsOF: uuid and name are only used as a pair. Can we consolidate as something like
`const mainWindowIdentity = {uuid: app.uuid, name: app.uuid}`
and change the below to just
`sendToClient<...>(mainWindowIdentity, ...)`

MichaelMCoates:
All tests passed except for lint. Pushed up a lint commit. Will squash and merge now.

PR #268, opened 2019-02-08
[RUN-5010] Set no maxWidth/height to undefined instead of -1

pbaize:
There is a core issue where having any maxWidth/maxHeight prevents maximization. The updateOptions call should be called with undefined instead. This should hold you over until a proper runtime fix is released.

### DIFF
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -842,9 +842,9 @@ export class DesktopWindow implements DesktopEntity {
                         }
                     },
                     minWidth: resizeConstraints.x.minSize,
-                    maxWidth: resizeConstraints.x.maxSize === Number.MAX_SAFE_INTEGER ? undefined : resizeConstraints.x.maxSize,
+                    maxWidth: resizeConstraints.x.maxSize,
                     minHeight: resizeConstraints.y.minSize,
-                    maxHeight: resizeConstraints.y.maxSize === Number.MAX_SAFE_INTEGER ? undefined : resizeConstraints.y.maxSize,
+                    maxHeight: resizeConstraints.y.maxSize,
                 }));
             }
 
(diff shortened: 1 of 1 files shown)


ryanpetersonOF:
Will the proper runtime fix expect `-1` or `undefined`?

tomrobertsOF:
I tested this on Thursday and it seemed like setting undefined to start with was fine, bit if you set any value and then undefined the same issue occurred. I'll recheck once I'm on the office, but if so this workaround won't be feasible

pbaize:
@tomrobertsOF Good catch. Just using maxInt is technically the "correct" way to handle this, and the runtime works as expected with MaxInt. @ryanpetersonOF in light of this the runtime fix will probably just be either using MaxInt for everything and replacing -1 altogether, or supporting setting -1 TBD.

tomrobertsOF:
@pbaize, does setting to maxInt allow maximising the window? I vaguely remember it didn't when testing.

pbaize:
It does on my machine on .38. I think this might be an enableChromium change that happened in 37

tomrobertsOF:
Hmm, ok. I'll double-check on my side. Would be perfect if it did.

tomrobertsOF:
@pbaize, just tested with MAX_SAFE_INTEGER and see same behaviour as undefined: if set from nothing, it doesn't prevent maximising, but will not restore maximizability when maxWidth has been set to any other number. Are you seeing the same?

tomrobertsOF:
Will run this through CI just in case changing the numbers causes issues for tests.

MichaelMCoates:
Got some test failures in CI:
4 - `tabGroupResizeConstraints`
1 - `validateGroup`
1 - `updateTabProperties`

pjbroadbent:
Testing revealed some additional issues here. Replacing this PR with #280.

PR #269, opened 2019-02-08
[SERVICE-341] S&R accounts for hidden parent windows during restoration

MichaelMCoates:
Started with fixing SERVICE-341, but making changes there ended up taking me to different parts of S&R, so ended up making stability improvements to keep the tests running well. I also spilled over into SERVICE-310 in the process. 

- Made some changes to the end of restoration logic (Restore now waits for all placeholders to close before regrouping) 
- Added `autoShow` parameters to child creation to test restoration flickering.
- Changed up `positionWindow` logic. 
- Changed up placeholder logic.

### DIFF
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -155,7 +155,7 @@ export interface WorkspaceApp {
      * This is only present if the application was started programatically. For applications started from a manifest,
      * `manifestUrl` will be present instead.
      */
-    initialOptions?: {uuid: string, url: string};
+    initialOptions?: {uuid: string, url: string, mainWindowOptions: {autoShow: boolean}};
 
     /**
      * Application identifier
--- a/src/client/types.ts
+++ b/src/client/types.ts
@@ -155,7 +155,7 @@ export interface WorkspaceApp {
      * This is only present if the application was started programatically. For applications started from a manifest,
      * `manifestUrl` will be present instead.
      */
-    initialOptions?: {uuid: string, url: string, mainWindowOptions: {autoShow: boolean}};
+    initialOptions?: fin.ApplicationOptions;
 
     /**
      * Application identifier
--- a/src/provider/tabbing/TabService.ts
+++ b/src/provider/tabbing/TabService.ts
@@ -208,6 +208,7 @@ export class TabService {
                 await tabGroup.addTabs(tabs, groupDef.groupInfo.active);
                 await tabGroup.window.sync();
 
+                // TODO: Change to look at groupDef.groupInfo.state
                 if (tabGroup.state === 'maximized') {
                     await tabGroup.maximize();
                 } else if (tabGroup.state === 'minimized') {
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -11,7 +11,7 @@ import {promiseMap} from '../snapanddock/utils/async';
 
 import {SCHEMA_MAJOR_VERSION} from './create';
 import {regroupWorkspace} from './group';
-import {addToWindowObject, childWindowPlaceholderCheck, childWindowPlaceholderCheckRunningApp, createNormalPlaceholder, createTabbedPlaceholderAndRecord, inWindowObject, parseVersionString, positionWindow, SemVer, TabbedPlaceholders, wasCreatedProgrammatically, WindowObject} from './utils';
+import {addToWindowObject, childWindowPlaceholderCheck, childWindowPlaceholderCheckRunningApp, createNormalPlaceholder, createTabbedPlaceholderAndRecord, inWindowObject, parseVersionString, positionWindow, SemVer, TabbedPlaceholders, wasCreatedProgrammatically, WindowObject, waitUntilAllPlaceholdersClosed} from './utils';
 
 const appsToRestore = new Map();
 const appsCurrentlyRestoring = new Map();
@@ -232,21 +232,26 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
         }
     });
     // Wait for all apps to startup
-    const startupResponses = await Promise.all(startupApps);
-    // Consolidate application responses
-    const allAppResponses = apps.map(app => {
-        const appResponse = startupResponses.find(appRes => appRes.uuid === app.uuid);
-        return appResponse ? appResponse : app;
-    });
-    layout.apps = allAppResponses;
+    await Promise.all(startupApps);
+
+    // Wait for all child windows to appear. Continue and Warn if placeholders aren't closed in 30 seconds.
+    try {
+        await waitUntilAllPlaceholdersClosed();
+    } catch (error) {
+        console.warn(error);
+    }
+
     // Regroup the windows
-    await regroupWorkspace(allAppResponses).catch(console.log);
+    await regroupWorkspace(layout.apps).catch(console.log);
+
     // Validate groups
     for (const group of model.snapGroups) {
         group.validate();
     }
 
     apiHandler.sendToAll('workspace-restored', layout);
+
+    console.log("Restore completed");
 
     // Send the layout back to the requester of the restore
     return layout;
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -216,7 +216,7 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
                     await ofAppNotRunning.run().catch(console.log);
                     await model.expect({name, uuid});
 
-                    // If the app's main window is tabbed or mimnimized, position it in the proper location.
+                    // If the app's main window is tabbed or minimized, position it in the proper location.
                     if (app.mainWindow.isTabbed || app.mainWindow.state === 'minimized') {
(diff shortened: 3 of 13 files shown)


src/provider/workspaces/restore.ts
```diff
@@ -225,22 +232,27 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
         }
     });
     // Wait for all apps to startup
-    const startupResponses = await Promise.all(startupApps);
-    // Consolidate application responses
-    const allAppResponses = apps.map(app => {
-        const appResponse = startupResponses.find(appRes => appRes.uuid === app.uuid);
-        return appResponse ? appResponse : app;
-    });
-    layout.apps = allAppResponses;
```
bryangaleOF: Why is this no longer needed? Or was this not needed in the first place?

src/client/types.ts
```diff
@@ -155,7 +155,7 @@ export interface WorkspaceApp {
      * This is only present if the application was started programatically. For applications started from a manifest,
      * `manifestUrl` will be present instead.
      */
-    initialOptions?: {uuid: string, url: string};
+    initialOptions?: {uuid: string, url: string, mainWindowOptions: {autoShow: boolean}};
```
pjbroadbent: Can we use `fin.ApplicationOptions` here?

src/provider/tabbing/TabService.ts
```diff
@@ -208,6 +208,7 @@ export class TabService {
                 await tabGroup.addTabs(tabs, groupDef.groupInfo.active);
                 await tabGroup.window.sync();
 
+                // TODO: Change to look at groupDef.groupInfo.state
```
pjbroadbent: For tracking purposes: [SERVICE-323](https://appoji.jira.com/browse/SERVICE-323)

src/provider/workspaces/restore.ts
```diff
@@ -205,6 +205,13 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
                     // If application created programmatically
                     if (wasCreatedProgrammatically(app)) {
                         console.warn('App created programmatically, app may not restart again:', app);
+                        if (app.initialOptions) {
+                            if (app.initialOptions.mainWindowOptions) {
+                                app.initialOptions.mainWindowOptions.autoShow = false;
+                            } else {
+                                app.initialOptions.mainWindowOptions = {autoShow: false};
```
pjbroadbent: Why do we default to `false` here? Would using `app.mainWindow.isShowing` simplify the restore logic further down the line?..

src/provider/workspaces/restore.ts
MichaelMCoates: So originally, we were only regrouping/affecting windows that were not running. By regrouping all applications in a layout instead, we make sure that everything's grouped at the end of the day.

src/client/types.ts
MichaelMCoates: Changing to fin.WindowOptions

MichaelMCoates: Changed

src/provider/tabbing/TabService.ts
MichaelMCoates: Thanks!

src/provider/workspaces/restore.ts
MichaelMCoates: So this was a UX change. Ideally, when we restore, we want the application to only appear in its restored spot. If `autoShow` is `true`, the application will first appear in its default location before moving over to its restore location. By overriding it to `false`, we can make programmatic applications show when and where we want them to.

MichaelMCoates: I could think about going back to a `show-requested` listener during placeholder generation instead? That way this works for manifest-created windows.

src/provider/workspaces/restore.ts
bryangaleOF: Were we only regrouping/affecting windows that were not running? My reasoning on this was that allAppResponses would contain a WorkspaceApp for every app, regardless of whether it was already running or not. The big promiseMap above this will always give something for each app, running or not, and the deleted apps.map(...) won't remove anything, just replace potentially out-of-date WorkspaceApp objects

bryangaleOF: My understanding of this change is that we keep working with the WorkspaceApp objects that came in originally with the call to restoreWorkspace, where before we switched to working with the WorkspaceApp objects that came back from each app's WorkspaceAPI.RESTORE_HANDLER. I just wanted to confirm that's fine

MichaelMCoates: Oh wait, yes, you're completely right. Totally missed the return ternary. Reverting, thanks!

src/provider/workspaces/restore.ts
MichaelMCoates: Went back to `show-requested` for normal placeholders and `shown` for tab placeholders.

src/provider/workspaces/restore.ts
```diff
@@ -213,8 +214,20 @@ export const restoreWorkspace = async(payload: Workspace, identity: Identity): P
...
+                        await positionWindow(app.mainWindow);
+                    }
+
+                    if (app.initialOptions && app.initialOptions.mainWindowOptions) {
+                        // If the app's autoShow is set to false, but isShowing is true for the application, then position it.
+                        // Otherwise, placeholder resolution may not complete properly.
+                        if (app.initialOptions.mainWindowOptions.autoShow === false && app.mainWindow.isShowing) {
+                            await positionWindow(app.mainWindow);
+                        }
+                    }
```
bryangaleOF: Why is it only in these specific cases we want to call positionWindow(), and not do it in other cases?

MichaelMCoates: My midnight thought process was: We want to limit our `positionWindow` calls as much as possible, to make sure that we don't have situations in which the app `autoShow`s, which triggers the placeholder `positionWindow`, then this `positionWindow` triggers and we get into a weird state. Then again, it seems like the opposite case (a window that is expected to show itself but doesn't, freezing the restoration) might be worse, so maybe it's better to call `positionWindow` on everything? I'll play around with it.

src/provider/workspaces/utils.ts
```diff
@@ -14,44 +14,80 @@ export interface SemVer {
...
+    if (win.name.startsWith('Placeholder-')) {
+        numOfPlaceholders--;
+        if (numOfPlaceholders === 0 && placeholderResolve) {
+            placeholderResolve();
+            placeholderResolve = undefined;
+        }
+    }
+});
+
+export async function waitUntilAllPlaceholdersClosed() {
```
bryangaleOF: Think this misbehaves if ever called a second time before the timeout has elapsed. Could placeholderResolve be an array of  (() => void)s? Or could this warn/throw an error if called a second time before resolving the first time?

src/provider/workspaces/utils.ts
```diff
@@ -14,44 +14,80 @@ export interface SemVer {
     patch: number;
 }
 
+// TODO: Create Placeholder and PlaceholderStore classes?
+// This keeps track of how many placeholders we have open, so we know when we can start regrouping a layout.
+let numOfPlaceholders = 0;
+let placeholderResolve: (() => void)|undefined;
```
bryangaleOF: Is there a more descriptive name to indicate it's to do with waiting for all placeholders to close?

src/provider/workspaces/utils.ts
```diff
@@ -64,48 +100,65 @@ export const createNormalPlaceholder = async (win: WorkspaceWindow) => {
...
     };
+    // We add a listener to show-requested so that the window shows up in the location it's supposed to be restored at.
+    // If we added a listener to shown, we'd see the window appear in its original spot and then flash to the next spot.
     await actualWindow.addListener('show-requested', updateOptionsAndShow);
-
-    return placeholder;
+    return placeholderWindow;
 };
 
+async function delay(milliseconds: number) {
```
bryangaleOF: Are we following a convention of exported functions first? Analogous to ordering a class public->protected->private in other languages

src/provider/workspaces/restore.ts
MichaelMCoates: Reverted to just `positionWindow`. Didn't see any weird behavior, but keep a lookout for strange positioning or grouping behavior.

src/provider/workspaces/utils.ts
MichaelMCoates: Changed to `functionToContinueRestorationWhenPlaceholdersClosed`

src/provider/workspaces/utils.ts
MichaelMCoates: Firmed up the logic around this. Throws an error if in middle of a restoration, and resets the state of the placeholder tracker once it resolves or rejects.

src/provider/workspaces/utils.ts
MichaelMCoates: Rearranged the file to follow the pattern. Removed functions that were no longer being used.

MichaelMCoates:
Passes CI. Merge?

PR #272, opened 2019-02-12
[SERVICE-292] Generate & publish config docs

bryangaleOF:
This PR uses bootprint to generate JSON schema documentation as the least bad solution available. As with all Node-based JSON documentation schema generators, bootprint is flagged as depending on packages with security vulnerabilities, hence we do not want to include it in our devDependencies, hence Jenkins having the extra step of installing bootprint.

Also, all schema files have been consolidated into a single file, due to unreliable reference handling by bootprint. This PR also adds a further reference to 'Scope' in layouts-config so that it is included in the generated documentation for layouts-config.


package.json
```diff
@@ -53,6 +55,10 @@
     "webpack-cli": "^3.1.0",
     "webpack-dev-middleware": "^3.4.0"
   },
+  "peerDependencies": {
+    "bootprint": "^2.0.1",
+    "bootprint-json-schema": "2.0.0-rc.3"
+  },
```
pjbroadbent: Seems NPM warns about any uninstalled peer dependencies - which we'd want to avoid.

Stick with current setup, but remove the `peerDependencies` section.

PR #275, opened 2019-02-13
[SERVICE-323] generateLayout has incorrect state for maximized windows

bryangaleOF:
Leftover fix that I'd been holding off on until this file was free. Tab windows now respect maximized/minimized state when restoring

Waiting on Jenkins at 04156f3


tomrobertsOF:
Was it `switchTab` or `swapTab` that was causing the problems? I would have thought swapTab would also need some kind of handling, but may well be wrong

bryangaleOF:
> Was it `switchTab` or `swapTab` that was causing the problems? I would have thought swapTab would also need some kind of handling, but may well be wrong

swapTab calls switchTab, and there's a less impactful fix only changing swapTab, but this felt like a neater solution rather than peppering the class with more applyProperties calls

tomrobertsOF:
> > Was it `switchTab` or `swapTab` that was causing the problems? I would have thought swapTab would also need some kind of handling, but may well be wrong
> 
> swapTab calls switchTab, and there's a less impactful fix only changing swapTab, but this felt like a neater solution rather than peppering the class with more applyProperties calls

Ok, sounds sensible. Wanted to make sure we hadn't missed some other edge case while we were in there.

bryangaleOF:
I've spotted another problem, occurs specifically when restoring a minimized tab group, when the app is already running. Will address as part of this PR

bryangaleOF:
Submitting this for review with minimization state restoration commented out until SERVICE-351 is fixed. Think is still worth getting into develop as is rather than scope-creeping this ticket.

The bulk of this PR is DesktopTabWindow firing state change events, and the boilerplate needed to support that. The tabstrip now listens to tab-group-maximized and tab-group-minimized events to change the maximize/restore button

Also, when saving a maximized window, we save the pre-maximization dimensions, so that the restore button works correctly post-workspace restore

Currently waiting on Jenkins

bryangaleOF:
Tests passed on Jenkins

src/provider/model/DesktopTabGroup.ts
```diff
@@ -219,28 +218,39 @@ export class DesktopTabGroup implements DesktopEntity {
...
     /**
      * Restores the tab set window.  If the tab set window is in a maximized state we will restore the window to its "before maximized" bounds.
      */
     public async restore(): Promise<void> {
-        if (this._isMaximized) {
-            if (await this.activeTab.currentState.state === 'minimized') {
-                await Promise.all(this._tabs.map(tab => tab.applyProperties({state: 'normal'})));
-            } else if (this._beforeMaximizeBounds) {
-                this._isMaximized = false;
+        if (await this.state === 'minimized') {
```
tomrobertsOF: Why is this awaited? Isn't it a synchronous getter?

src/provider/model/DesktopTabGroup.ts
```diff
@@ -219,28 +218,39 @@ export class DesktopTabGroup implements DesktopEntity {
...
-                    center: {x: bounds.center.x, y: bounds.center.y - bounds.halfSize.y - (this._config.height / 2)},
-                    halfSize: {x: bounds.halfSize.x, y: this._config.height / 2}
-                });
-                await this.activeTab.applyProperties(bounds);
-            }
+            return result;
         } else {
-            await Promise.all(this._tabs.map(tab => tab.applyProperties({state: 'normal'})));
+            if (this._isMaximized) {
+                if (await this.activeTab.currentState.state === 'minimized') {
```
tomrobertsOF: As above, currentState is a synchronous getter. No await should be needed

src/provider/model/DesktopTabGroup.ts
```diff
@@ -219,28 +218,39 @@ export class DesktopTabGroup implements DesktopEntity {
...
+                    await Promise.all(this._tabs.map(tab => tab.applyProperties({state: 'normal'})));
+                } else if (this._beforeMaximizeBounds) {
+                    this._isMaximized = false;
+
+                    const bounds: Rectangle = this._beforeMaximizeBounds;
```
tomrobertsOF: This will be assigned by reference, so you're editing the original object. If that's what you want, then no need to reassign, otherwise should clone it instead, e.g.
```
const newBounds: Rectangle = {...this._beforeMaximizeBounds}
```

src/provider/model/DesktopTabGroup.ts
```diff
@@ -420,19 +432,21 @@ export class DesktopTabGroup implements DesktopEntity {
             const redrawRequired: boolean = prevTab && !RectUtils.isEqual(tab.currentState, this._activeTab.currentState);
             this._activeTab = tab;
 
+            const tabState = this.state === 'minimized' ? 'minimized' : 'normal';
```
tomrobertsOF: Is the handling for maximized the same as for normal in all cases?

src/provider/model/DesktopTabGroup.ts
```diff
@@ -462,6 +476,29 @@ export class DesktopTabGroup implements DesktopEntity {
         this._validateGroup.call();
     }
 
+    public getSaveDimensions(): TabGroupDimensions {
```
tomrobertsOF: Is there any particular reason why we use top-left for this when the rest of the service tends to use centre-halfSize?

src/provider/tabbing/tabstrip/main.ts
```diff
@@ -41,6 +42,20 @@ const createLayoutsEventListeners = () => {
             if (props.title) tab.updateText(props.title);
         }
     });
+
+    const maximizeElem: HTMLElement|null = document.getElementById('window-button-maximize');
```
tomrobertsOF: Pedantic, but `maximizeButton` may be more descriptive?

src/provider/tabbing/tabstrip/main.ts
```diff
@@ -41,6 +42,20 @@ const createLayoutsEventListeners = () => {
...
+
+    layouts.tabstrip.addEventListener('tab-group-maximized', (event: TabGroupMaximizedEvent) => {
+        tabManager.isMaximized = true;
+        maximizeElem!.classList.add('restore');
```
tomrobertsOF: I'm never a huge fan of `!` if it can be avoided. Could we maybe just wrap this in an `if (maximizeElem) {...}`?

src/provider/workspaces/utils.ts
```diff
@@ -47,15 +47,19 @@ export async function waitUntilAllPlaceholdersClosed() {
...
     try {
         const {isShowing, isTabbed} = win;
 
         const ofWin = await fin.Window.wrap(win);
         await ofWin.setBounds(win);
 
         if (isTabbed) {
-            await ofWin.show();
+            if (replacingPlaceholder) {
+                // When not replacing a placeholder, and hence no further handling done in a show-requested listener,
```
tomrobertsOF: I had a lot of trouble figuring out what this comment was actually saying. Might be just me being dense, but could you see if you can make it a bit clearer?

src/provider/model/DesktopTabGroup.ts
bryangaleOF: Hmm, brought in with code moved from APIHandler. Will fix.

src/provider/tabbing/tabstrip/main.ts
bryangaleOF: Inclined to keep as is. We're using Elem for HTML elements elsewhere, and I'm more interested at this point in it's HTML element nature than its button nature

src/provider/tabbing/tabstrip/main.ts
bryangaleOF: Another one I'll push back on. `!` for dealing the the result of `getElementById()` seems to be pretty standard in layouts, and indicates that we don't reasonably expect it to be null

src/provider/model/DesktopTabGroup.ts
bryangaleOF: I assume the assignment of `this._beforeMaximizeBounds` to the local `bounds` was done mostly for brevity. Otherwise, the proceeding lines get hard to read. There's no editing of _beforeMaximizeBounds. The only scary bit might be passing `bounds` into applyProperties, but following that down, that doesn't change the object passed in at all, and it's a line that's been there a while. Plus nothing anywhere edits the object at `this._beforeMaximizeBounds`, only reassigns, so there's no danger of weird race conditions

src/provider/workspaces/utils.ts
bryangaleOF: Yeah, this was really obtuse. Fixing.

src/provider/model/DesktopTabGroup.ts
bryangaleOF: In this case yes, since tabs are never maximized from a internal Windows PoV. The only time we don't want tabState to be normal is when the tab group is minimized. Otherwise from a low-level windows point of view, the tab will be normal.

src/provider/model/DesktopTabGroup.ts
bryangaleOF: I was wondering that myself! Will check with @pjbroadbent about this, but feel like any change in save format would be it's own PR

bryangaleOF: @pjbroadbent able to weigh in?

src/provider/model/DesktopTabGroup.ts
tomrobertsOF: Ah, I must have misread this on first glance. Though we were assigning to bounds below, but apparently not.

src/provider/model/DesktopTabGroup.ts
tomrobertsOF: Cool. As long as this was intentional.

src/provider/tabbing/tabstrip/main.ts
tomrobertsOF: Ok

src/provider/model/DesktopTabGroup.ts
pjbroadbent: Anything that was refactored to use `DesktopModel` was made consistent and now uses `centre`/`halfSize`. We never got round to refactoring save & restore in the same way, so much of the workspaces code is still using l/t/r/b or l/t/w/h.

Would be good to align, but we should also stick to OpenFin API conventions (l/t/r/b/w/h) for anything that an external app is expected to handle. Any kind of change would likely touch a lot of the code, should handle as future tech debt rather than as the current release.

src/provider/tabbing/tabstrip/main.ts
pjbroadbent: I would put the `!` on the declaration though, to avoid having to place it on every `maximizeElem` usage.

e.g:
`const maximizeElem: HTMLElement = document.getElementById('window-button-maximize')!;`

PR #277, opened 2019-02-16
[SERVICE-371] Update tabstrip API Docs

ryanpetersonOF:
**Do Not Merge Yet**

Updates the tabstrip client API and some related internal provider logic.

API Changes:
* `endDrag(event: DragEvent, window: Identity)` is now `endDrag()`
No longer takes any parameters.  We already know `window` from `startDrag` and `event` is handled in the provider.


src/provider/APIHandler.ts
```diff
@@ -282,8 +282,12 @@ export class APIHandler {
         }
     }
 
-    private reorderTabs(newOrdering: WindowIdentity[], tabId: ProviderIdentity): void {
-        const tab: DesktopWindow|null = this._model.getWindow(tabId as WindowIdentity);
+    private reorderTabs(newOrdering: WindowIdentity[], tabstrip: ProviderIdentity): void {
+        if (Array.isArray(newOrdering) && (!newOrdering || newOrdering.length === 0)) {
```
bryangaleOF: Will the isArray check ever be false? And if it can ever be false, I'm not sure about the logic here - 'if it's not an array, don't do any validation checks, but let it fail later rather than sooner'.

bryangaleOF: Worth renaming reOrderTabArray to reorderTabArray while you're here?

ryanpetersonOF: Fixed both comments - good catch on the logic!

src/provider/APIHandler.ts
```diff
@@ -305,17 +309,18 @@ export class APIHandler {
         }
     }
 
-    private startDrag(payload: {window: WindowIdentity}, source: ProviderIdentity): void {
+    private startDrag(payload: WindowIdentity, source: ProviderIdentity): void {
```
tomrobertsOF: Now that we're not passing it as an object, we could probably change the param name to something more descriptive.

src/provider/APIHandler.ts
```diff
@@ -327,20 +332,7 @@ export class APIHandler {
...
-            console.error('Window is not registered for tabbing');
-            throw new Error('Window is not registered for tabbing');
-        }
-
-        const target = this._tabService.getTarget(tab);
+    private async endDrag(): Promise<void> {
         this._tabService.dragWindowManager.hideWindow();
-
-        if (target) {
-            await this._tabService.applyTabTarget(target);
```
tomrobertsOF: Is this really not needed any more? How does the drop trigger the change now?

ryanpetersonOF: Correct.  The drag window captures the drop event and then emits a signal.  The WindowHandler answers then calls applyTabTarget.  

The only reason at this point that we even need `endDrag` is in the case that the user drops the tab on the source tab strip, which is on top of the drag window.  In this case we cant capture the drop event and this call is required to hide the drag window.

src/provider/APIHandler.ts
ryanpetersonOF: Fixed.  Renamed to `identity`

src/provider/APIHandler.ts
tomrobertsOF: Cool. Sounds good.

src/provider/WindowHandler.ts
```diff
@@ -86,12 +86,17 @@ export class WindowHandler {
...
         this.view.update(target);
     }
 
-    private onTabDrop() {
+    private onTabDrop(window: DesktopWindow) {
         this.view.update(null);
+
+        const target = tabService.getTarget(window);
+
+        if (target) {
```
pjbroadbent: Need a `target.valid` check here? Or is that done within the apply method?

src/provider/tabbing/DragWindowManager.ts
```diff
@@ -22,11 +23,18 @@ export class DragWindowManager {
      *
      * Arguments: None.
```
pjbroadbent: Update docs

src/provider/tabbing/DragWindowManager.ts
```diff
@@ -22,11 +23,18 @@ export class DragWindowManager {
...
     // Usecase: failsafe for the drag window overlay should somehow it not close, the user would be "locked out" of the desktop.
-    private _hideTimeout: number|NodeJS.Timer;
+    private _hideTimer: number|NodeJS.Timer;
+
+    /**
+     * Timeout for the drag window failsafe.
+     * Consideration must be taken for the case of user dragging tab on top of the source tabstrip - Timer is not cleared or reset as no event from us is
+     * generated.
+     */
+    private readonly HIDE_TIMEOUT = 30000;
```
pjbroadbent: Should our fail-safe be adding a `click` listener to the overlay? To avoid an actual tab operation from timing-out?..

May just be trading one set of edge cases for another..?

src/provider/tabbing/DragWindowManager.ts
```diff
@@ -65,27 +73,34 @@ export class DragWindowManager {
...
+        this.setHideTimer();
     }
 
     /**
      * Hides the drag window overlay.
      */
     public hideWindow(): void {
-        DragWindowManager.onDragDrop.emit();
-        this._window.hide();
+        // Check if we've got a timer running.  If not then we're liking being called from an invalid drag end event (Erroneous or duplicate call)
```
pjbroadbent: `likely`?..

src/provider/tabbing/DragWindowManager.ts
pjbroadbent: Make this a static? Think that's what we do with all other constants.

src/client/tabstrip.ts
```diff
@@ -4,48 +4,67 @@
...
+ * Only one dragging operation should ever be taking place.
+ *
+ * ```ts
+ * import {tabstrip} from 'openfin-layouts';
+ *
+ * tabstrip.startDrag({uuid: 'App0', name: 'App0'});
+ * ```
+ *
+ * @param identity: The identity of the tab which is being dragged.
+ * @throws `Promise.reject`: If `identity` is not a valid {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity}.
```
pjbroadbent: We throw an `Error`, rather than a `Promise.reject`?..

Not sure what the proper way to document these exceptions is. Is this how we do it in the rest of the API?

src/client/tabstrip.ts
```diff
@@ -4,48 +4,67 @@
...
- * Ends the HTML5 Dragging Sequence.
+ * Informs the layouts service a tab HTML5 drag sequence has ended.  Required at the end of any tabstrip drag operation.
+ * Only one dragging operation should ever be taking place.
+ *
+ * ```ts
+ * import {tabstrip} from 'openfin-layouts';
+ *
+ * window.document.body.addEventListener("dragend", (event) => {
+ *      tabstrip.endDrag();
+ * })
```
pjbroadbent: `;` :-)

Can you add a `.addEventListener("dragstart")` example to the `startDrag` call too? Gives a better idea of usage.

src/client/tabstrip.ts
ryanpetersonOF: This is how its done in the rest of the API.  If an error is thrown it will return a reject on the client side.

src/provider/WindowHandler.ts
ryanpetersonOF: Fixed.  Added validity check, as seen in other places of the same file.

src/provider/tabbing/DragWindowManager.ts
ryanpetersonOF: Fixed

src/provider/tabbing/DragWindowManager.ts
ryanpetersonOF: Replaced the whole hideTimer with the click event.  Much better :)

src/provider/tabbing/DragWindowManager.ts
ryanpetersonOF: Thanks, fixed

src/client/tabstrip.ts
ryanpetersonOF: Example added. Good call

PR #261, opened 2019-02-06
[SERVICE-320] Maximising a tabbed window from the OS-level can cause issues for the tabSet

bryangaleOF:
Use updateState on DesktopWindow to make window non-maximizable when added to tab group

PR #282, opened 2019-02-18
[SERVICE-348][SERVICE-367] Block maximize actions on tabgroups that contain a window with maximizable: false or maxWidth/maxHeight constraints

tomrobertsOF:
- TabGroups will check their maximizability before maximising and throw an error if false.
- `updateGroupConstraints` will now also update the maximizability of the tabGroup.
- TabGroup maximizability is determined by the union of all windows' maximizability and whether they have any maximum height/width constraints.

### DIFF
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -203,7 +203,26 @@ export class DesktopTabGroup implements DesktopEntity {
      * Maximizes the tab set window.  This will resize the tab window to as large as possible with the tab set window on top.
      */
     public async maximize(): Promise<void> {
-        if (!this._isMaximized) {
+        if (!this.currentState.maximizable) {
+            // This all looks a bit messy, but is basically just determining which windows are preventing the maximize and including them in the error message.
+            const nonMaximizableWindows: string[] = this._tabs.filter(tab => !tab.applicationState.maximizable).map(tab => tab.id);
+            if (nonMaximizableWindows.length > 0) {
+                throw new Error(`Unable to maximize tabGroup: The following tas are not resizable: [${nonMaximizableWindows.join(', ')}]`);
+            }
+            const sizeConstrainedWindows = this._tabs
+                                               .filter(
+                                                   tab => tab.applicationState.resizeConstraints.x.maxSize < Number.MAX_SAFE_INTEGER ||
+                                                       tab.applicationState.resizeConstraints.y.maxSize < Number.MAX_SAFE_INTEGER)
+                                               .map(tab => tab.id);
+            if (sizeConstrainedWindows.length > 0) {
+                throw new Error(`Unable to maximize tabGroup: The following tabs have maximum size constraints: [${sizeConstrainedWindows.join(', ')}]`);
+            }
+
+            // This should never be called since currentState.maximizable should only be false if one of
+            // the two conditions above is met. This catch-all is here just in case something weird happens.
+            throw new Error('Unable to maximize tabGroup: Group is not maximizable');
+        }
+        if (!this._isMaximized && this.currentState.maximizable) {
             // Before doing anything else we will undock the tabGroup (mitigation for SERVICE-314)
             if (this.snapGroup.entities.length > 1) {
                 await this.setSnapGroup(new DesktopSnapGroup());
@@ -656,6 +675,10 @@ export class DesktopTabGroup implements DesktopEntity {
         result.y.resizableMin = false;  // Cannot resize on the edge between tab and tabstrip (SERVICE-287)
         // Apply the new constraints to all windows
         await Promise.all(this.tabs.map((tab: DesktopWindow) => tab.applyProperties({resizeConstraints: result})));
+
+        // Changes to constraints also affect the maximizability of the tabgroup, so we update that here too
+        this._groupState.maximizable = this._tabs.every(tab => tab.applicationState.maximizable) &&        // All tabs must be maximizable
+            result.x.maxSize === Number.MAX_SAFE_INTEGER && result.y.maxSize === Number.MAX_SAFE_INTEGER;  // No tabs have maxSize constraints
     }
 
     // Will check that all of the tabs and the tabstrip are still in the correct relative positions, and if not
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -207,7 +207,7 @@ export class DesktopTabGroup implements DesktopEntity {
             // This all looks a bit messy, but is basically just determining which windows are preventing the maximize and including them in the error message.
             const nonMaximizableWindows: string[] = this._tabs.filter(tab => !tab.applicationState.maximizable).map(tab => tab.id);
             if (nonMaximizableWindows.length > 0) {
-                throw new Error(`Unable to maximize tabGroup: The following tas are not resizable: [${nonMaximizableWindows.join(', ')}]`);
+                throw new Error(`Unable to maximize tabGroup: The following tabs are not resizable: [${nonMaximizableWindows.join(', ')}]`);
             }
             const sizeConstrainedWindows = this._tabs
                                                .filter(
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -204,25 +204,24 @@ export class DesktopTabGroup implements DesktopEntity {
      */
     public async maximize(): Promise<void> {
         if (!this.currentState.maximizable) {
-            // This all looks a bit messy, but is basically just determining which windows are preventing the maximize and including them in the error message.
             const nonMaximizableWindows: string[] = this._tabs.filter(tab => !tab.applicationState.maximizable).map(tab => tab.id);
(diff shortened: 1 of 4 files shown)


bryangaleOF:
I'm concerned that throwing an exception isn't the right thing to do. The exception in this case doesn't indicate something exceptional has gone wrong, plus will stomp normal program flow unnecessarily. I feel like even silently doing nothing would be preferable.

More minor, I'd probably pull out something like getNonMaximizable tabs and getSizeConstrainedTabs, which would save duplicating knowledge in two spread apart parts of the file, and may mean less commenting is needed. I'd also move the map(tab => tab.id) into where the error string is created, so error string creation is separated from the check itself.

tomrobertsOF:
The exception propagates back to the client which made the `tabbing.minimize()` call. If we fail silently the tabstrip will have no way of knowing whether the group actually maximized or not.

src/provider/model/DesktopTabGroup.ts
```diff
@@ -203,7 +203,26 @@ export class DesktopTabGroup implements DesktopEntity {
      * Maximizes the tab set window.  This will resize the tab window to as large as possible with the tab set window on top.
      */
     public async maximize(): Promise<void> {
-        if (!this._isMaximized) {
+        if (!this.currentState.maximizable) {
+            // This all looks a bit messy, but is basically just determining which windows are preventing the maximize and including them in the error message.
```
pjbroadbent: Is a bit, but can be cleaned up: (?)
```
if (!this.currentState.maximizable) {
    const nonMaximizableWindows: string[] = this._tabs.filter(tab => !tab.applicationState.maximizable).map(tab => tab.id);
    const sizeConstrainedWindows: string[] = this._tabs.filter(tab => {
        const constraints: Point<ResizeConstraint> = tab.applicationState.resizeConstraints;
        return constraints.x.maxSize < Number.MAX_SAFE_INTEGER || constraints.y.maxSize < Number.MAX_SAFE_INTEGER;
    }).map(tab => tab.id);

    if (nonMaximizableWindows.length > 0) {
        throw new Error(`Unable to maximize tabGroup: The following tabs are not resizable: [${nonMaximizableWindows.join(', ')}]`);
    } else if (sizeConstrainedWindows.length > 0) {
        throw new Error(`Unable to maximize tabGroup: The following tabs have maximum size constraints: [${sizeConstrainedWindows.join(', ')}]`);
    } else {
        throw new Error('Unable to maximize tabGroup: Group is not maximizable');
    }
} else if (!this._isMaximized) {
    // . . .
}
```

bryangaleOF:
Are we sure we're not going to get an exception when it's called in TabService.createTabGroupsFromWorkspace? An exception here would disrupt the rest of the restore process. A possible solution would be an isMaximizable call (could also be useful down the line to expose to the tabstrip, so it can hide the maximize button?), or just wrapping those calls in try-catch blocks.

The tabstrip has the option of listening for maximized and minimized events, so doesn't necessarily need to get an exception to handle the window not maximizing.

I'd advocate the maximize to return a boolean instead of throw an exception (failure can happen as part of normal operation, nothing particularly breaks if something calls maximize but the window doesn't maximize, and the possibility of failure is indicated in the the method signature), but my only real blocker for approval on this is the possibility of it stomping the save and restore process

pjbroadbent:
IMO, having `maximize` throw in this sense makes sense - every API call has expected pre- and post-conditions. If you call `maximize` and the tab group isn't in the expected state when the promise completes, then in my mind that's the definition of when to throw an exception.

At the same time - if this function can now throw, we definitely need to make sure we're handling (even if "handling" just means "suppressing") those exceptions in any place that `maximize` is called within the provider. Whilst this is a "severe" enough of a failure to throw in the context of the client calling the `maximize` API, I feel it would be excessive to cause an entire `restore` call to fail - and not sure where this happens in the restore flow, but we certainly shouldn't fail in some broken, "part-restored" state.

pjbroadbent:
Can we add an integration test or two for this?

finos-admin:
Thank you for your pull request and welcome to our community! We require contributors to sign a [Contributor License Agreement](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Legal+Requirements#LegalRequirements-ContributorLicenseAgreement) and we don't seem to have CLAs on file for these contributors to the Pull Request: (@trobs757). In order for your PR to be reviewed and merged, please follow the directions at the link above.

Project team: please **do not merge this Pull Request** until Foundation staff have confirmed that a CLA is in place for the new contributor(s) listed above.

If there are any questions, please don't hesitate to [get in touch with our Infrastructure Support team](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/77955298/Getting+in+Touch#GettinginTouch-InfrastructureSupport). /CC @finos-admin

src/provider/model/DesktopTabGroup.ts
tomrobertsOF: Is that really much better?

finos-admin:
Thank you for your pull request and welcome to our community! We require contributors to sign a [Contributor License Agreement](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Legal+Requirements#LegalRequirements-ContributorLicenseAgreement) and we don't seem to have CLAs on file for these contributors to the Pull Request: (@trobs757). In order for your PR to be reviewed and merged, please follow the directions at the link above.

Project team: please **do not merge this Pull Request** until Foundation staff have confirmed that a CLA is in place for the new contributor(s) listed above.

If there are any questions, please don't hesitate to [get in touch with our Infrastructure Support team](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/77955298/Getting+in+Touch#GettinginTouch-InfrastructureSupport). /CC @finos-admin

tomrobertsOF:
Again, ignore the CLA error. I'm really struggling with git tonight.

finos-admin:
Thank you for your pull request and welcome to our community! We require contributors to sign a [Contributor License Agreement](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/75530375/Legal+Requirements#LegalRequirements-ContributorLicenseAgreement) and we don't seem to have CLAs on file for these contributors to the Pull Request: (@trobs757). In order for your PR to be reviewed and merged, please follow the directions at the link above.

Project team: please **do not merge this Pull Request** until Foundation staff have confirmed that a CLA is in place for the new contributor(s) listed above.

If there are any questions, please don't hesitate to [get in touch with our Infrastructure Support team](https://finosfoundation.atlassian.net/wiki/spaces/FINOS/pages/77955298/Getting+in+Touch#GettinginTouch-InfrastructureSupport). /CC @finos-admin

test/demo/tabbing/tabGroupResizeConstraints.test.ts
```diff
@@ -113,6 +114,31 @@ const defaultConstraints: Required<Constraints> = {
     }
 };
 
+testParameterized(
+    `Cannot maximize tabset when tab has maxWidth/Height`,
```
pjbroadbent: Can we also assert each window goes back to "normal" (the tab's previous constraints, or lack thereof) once de-tabbed?

src/provider/tabbing/TabService.ts
```diff
@@ -209,7 +209,7 @@ export class TabService {
 
                 // TODO: Change to look at groupDef.groupInfo.state
                 if (tabGroup.state === 'maximized') {
-                    await tabGroup.maximize();
+                    tabGroup.maximize().catch(console.warn);
```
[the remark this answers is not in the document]
tomrobertsOF: I don't remember removing those, they may have come from some merges. All I did was add the console.warn.

test/demo/tabbing/tabGroupResizeConstraints.test.ts
tomrobertsOF: We already have several tests that check propagation and restoration of window constraints when tabbing. This test is purely to check the maximizability of the tabGroup as a whole.

src/provider/tabbing/TabService.ts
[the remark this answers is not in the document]
tomrobertsOF: Are you looking at the current up-to-date version of the PR, because I can see the await in there.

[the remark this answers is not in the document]
tomrobertsOF: https://github.com/HadoukenIO/layouts-service/blob/16e9a6c5af7e398e672eb36f719727b015b774f0/src/provider/tabbing/TabService.ts#L199

test/demo/tabbing/tabGroupResizeConstraints.test.ts
pjbroadbent: Is maximizability covered in those existing tests, or only min/max sizes and `resizable`?

tomrobertsOF: We aren't touching maximizability of the windows at all in this test or this PR at all. That was added and tested for in https://github.com/HadoukenIO/layouts-service/pull/261/files

tomrobertsOF: If you think those tests aren't sufficient we can look at that, but I'd assert that that is well outside the scope of this story.

PR #283, opened 2019-02-18
[SERVICE-352] Prevent tabbing to maximized, un-tabbed windows

tomrobertsOF:
Changed `DesktopWindow.isActive` to also exclude maximized windows.

NOTE: This will lead to a known issue where it is possible to snap/tab to windows behind the maximized window.  After some discussions, that is preferable to tabbing/snapping to the maximized window's original position, and as the effort to resolve would be fairly substantial, it is being left to post-1.0.

### DIFF
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -425,7 +425,7 @@ export class DesktopWindow implements DesktopEntity {
 
     public get isActive(): boolean {
         const state: EntityState = this._currentState;
-        return !state.hidden && state.opacity > 0 && state.state !== 'minimized';
+        return !state.hidden && state.opacity > 0 && state.state === 'normal';
     }
 
     /**


bryangaleOF:
I feel like this change makes the name 'isActive' not make sense any more, and has the knock-on-effect of getWindowAt not behaving as expected from the name. This change works, but we're relying on the luck of isActive only being used by getWindowAt, and the luck of getWindowAt only being used by TabService.getTarget.

I feel like it'd make sense to have the 'not maximized' check in TabServices, as that feels like a good place to hold the knowledge of what can and can't be a tab target

src/provider/model/DesktopWindow.ts
```diff
@@ -425,7 +425,7 @@ export class DesktopWindow implements DesktopEntity {
 
     public get isActive(): boolean {
```
pjbroadbent: We should likely clarify the meaning of "active" through a comment. It's gone from a fairly simple "is this user-visible" to something more nuanced.

This is something that may well become feature-dependent in the future, at which point this may need to revert back, and we do the feature-specific filtering at a higher level.

pjbroadbent:
Bryan raises a good point. Pushing the test down to this level may have unintended side-effects once we run the filtered windows through z-indexer. A maximised window is still visible, and should block any action that applies only to the top-most window. This bounds issue means we can't (easily) implement a full solution, but could still at least properly handle z-index in a subset of cases.

Just a question of if this really matters.. 😃

pjbroadbent:
Can we add an integration test for this?

tomrobertsOF:
This seems like it needs some re-thinking. Will close the PR and reopen once I have a better way of handling it.

PR #288, opened 2019-02-20
[SERVICE-342] Only enable functionality for 'service-aware' applications

pjbroadbent:
The service will only act upon "service-aware" applications - these are applications that explicitly declare the service within their manifest.

In the case of programmatically-launched apps, the service will only act upon them if the closet manifest-launched app within the hierarchy is service-aware.

Save & Restore introduces some edge cases around programmatic apps, as after a restore all programmatically-launched apps will have `layouts-service` as their parent app, rather than the app that initially launched them. There is now some integration between workspaces and the config loader in order to handle this:
- When restoring a layout, workspaces will tell the loader to use it's set of parentUUID's over the parentUUID the loader would get by calling `getInfo()` on the app.
- When saving a layout, workspaces will first check if the loader has a cached parentUUID that reflects the "initial" parent of the app. This is so that saving a restored layout doesn't lose the original parentUUID information from the very first workspace generation.

**NOTE:**
This PR also moves around a fair amount of util code, to get the testbed app/window utils usable within integration tests, without duplicating the manifest generation code between demo app and test runner servers. Main bulk of the story changes are in `src/provider/config` and `src/provider/workspaces`.


src/provider/config/Loader.ts
```diff
@@ -79,30 +96,76 @@ export class Loader<T> {
...
+            if (parentUuid && (this._appState.hasOwnProperty(parentUuid) || !isManifest)) {
+                // Fetch parent state
+                let parentState: AppState|null = this.getAppState(parentUuid);
+                if (!parentState) {
+                    // Parent must have been a manifest-app that declared the service, but no custom config
+                    console.log(`Late-registering ${parentUuid} as service-aware (manifest-launched) app`);
+                    parentState = this.getOrCreateAppState(fin.Application.wrapSync({uuid: parentUuid}), true);
+                }
+
+                // App *may* be service-aware on it's own, but if it's state hasn't already been created by this point, then inerhit awareness from parent app
```
MichaelMCoates: *inherit

MichaelMCoates:
Is there a chance that we miss the initial `application-created` event for the application that launches the service?

MichaelMCoates:
Also one case to consider is:
- ApplicationA is `service-aware`
- `AppA` launches `AppB` programmatically (which inherits service awareness from `AppA`). 
- We generate a workspace, close all windows, and restore the workspace
- `AppB` is programmatically launched by the `layouts-service` as a part of restoration.

What's the best way to make sure `AppB` gets service awareness? What if we close `AppA`, does `AppB` lose `service-awareness`? What if we close `AppA`, but then generate/restore `AppB`?

src/provider/config/Loader.ts
```diff
@@ -40,17 +41,33 @@ interface AppState {
...
+     *
+     * For any applications that are not service-aware, some default configuration can also be specified. Here,
+     * "service-aware" is defined as either declaring the service within the application's manifest, or being a
+     * programmatically-launched application, that was launched by a "service-aware" application.
+     *
+     * @param store Store to which application-defined config will be placed
+     * @param serviceName The name of the service, plus aliases if appropriate
+     * @param defaultConfig Optionally, config to add to the store for any application that isn't service-aware
+     */
+    constructor(store: Store<T>, serviceName: string|string[], defaultConfig?: T) {
```
bryangaleOF: Super subjective, but I'd probably go with `serviceNames` to capture that it can be one or more

src/provider/config/Loader.ts
```diff
@@ -79,30 +96,76 @@ export class Loader<T> {
...
+                console.log(`Registering ${identity.uuid} as a child of ${parentState.scope.uuid}`);
+                state.parent = parentState;
+                parentState.children.push(state);
             }
 
-            if (parentUuid && this._appState.hasOwnProperty(parentUuid)) {
-                const state = this.getAppState(app);
-                const parent = this.getAppState(fin.Application.wrapSync({uuid: parentUuid}));
+            // Use default config for any apps that don't reference the service in any way
+            if (!isServiceAware && this._defaultConfig && identity.uuid !== fin.Application.me.uuid) {
```
bryangaleOF: Didn't we already early exit this method if identity.uuid === fin.Application.me.uuid?

pjbroadbent: Yep, removed the redundant check.

pjbroadbent:
> Is there a chance that we miss the initial `application-created` event for the application that launches the service?

@MichaelMCoates We may miss the event, but after adding the listener the loader will request all running apps and add them to the state tracking. Should catch everything...

> Also one case to consider is ...

@MichaelMCoates Had the same realisation... Pushed some changes yesterday to integrate the loader with Save & Restore - so that the loader will now use the parentUUID from the workspace JSON (if set) over what it gets from the `getInfo()` call.

Not a simple thing to integrate, as the two are quite disconnected and trying to keep the config stuff service-agnostic, so we can re-use it in other services. Any thoughts on implementation are welcome.

Likely still some edge cases around the parent app issue, but unless they're major issues will have to postpone to next release.

src/provider/workspaces/restore.ts
```diff
@@ -49,6 +49,14 @@ export const appReadyForRestore = async(uuid: string): Promise<void> => {
 export const restoreWorkspace = async(payload: Workspace): Promise<Workspace> => {
     console.log('Restoring workspace:', payload);
 
+    // Ensure the loader links any apps being programmatically restored with their parentUUID at time of workspace
```
bryangaleOF: Capitalization consistency

test/provider/defaultConfig.test.ts
```diff
@@ -36,81 +28,118 @@ test.afterEach.always(async (t: TestContext) => {
...
     // Ensure docking is now enabled
-    await dragSideToSide(children[1], 'left', children[0], 'right', {x: 10, y: 50});
-    await assertGrouped(t, children[0], children[1]);
+    await dragSideToSide(childWindows[1], 'left', childWindows[0], 'right', {x: 10, y: 50});
+    await assertGrouped(t, childWindows[0], childWindows[1]);
+
+    await Promise.all(childApps.map(app => app.close()));
+});
+
+test('Loader will override parentUuid\'s with data in workspace when building app hierarchy', async (t: TestContext) => {
```
bryangaleOF: Don't think you want an apostrophe?

test/provider/defaultConfig.test.ts
```diff
@@ -36,81 +28,118 @@ test.afterEach.always(async (t: TestContext) => {
...
+
+    // Ensure docking is still disabled
+    await dragSideToSide(childWindows[1], 'left', childWindows[0], 'right', {x: 10, y: 50});
+    await assertNotGrouped(childWindows[0], t);
+    await assertNotGrouped(childWindows[1], t);
+
+    await Promise.all(childApps.map(app => app.close()));
+});
+
+test('When saving a previously-restored workspace, the generated workspace will import parentUuid\'s from Loader', async (t: TestContext) => {
```
bryangaleOF: Ditto apostrophe

scripts/server/spawn.js
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Creates express-compatible middleware function to generate custom application manifests.
+ * 
+ * Differs from createAppJsonMiddleware (defined in server.js), as this spawns custom demo windows, rather than 
+ * re-writing existing demo/provider manifests.
+ */
+function createCustomManifestMiddleware() {
+    return async (req, res, next) => {
+        const defaultConfig = await readJsonFile(path.resolve('res/demo/app.json')).catch(next);
+        const {uuid, url, frame, defaultWidth, defaultHeight, realmName, enableMesh, runtime, useService, defaultCentered, defaultLeft, defaultTop, provider, config} = {
```
bryangaleOF: Should you keep consistency of ordering here?

src/demo/spawn.ts
```diff
@@ -242,6 +218,13 @@ function getUrl(options: WindowData): string {
     const urlQueryParams = options.queryArgs || {};
     const urlQueryKeys = Object.keys(urlQueryParams);
 
+    // Resolve relative URL's
```
bryangaleOF: Apostrophe

test/provider/defaultConfig.test.ts
```diff
@@ -0,0 +1,116 @@
...
+test.beforeEach(async (t: TestContext) => {
+    t.context.windows = [];
+});
+test.afterEach.always(async (t: TestContext) => {
+    await Promise.all(t.context.windows.map(win => win.close()));
+
+    await teardown(t);
+});
+
+// async function isWindowRegistered(identity: Identity): Promise<boolean> {
```
bryangaleOF: Does this need to stay in here?

PR #290, opened 2019-02-20
[SERVICE-350] Prevent child windows from snapping to other windows during restoration

MichaelMCoates:
- Reverse ordering of parent and child `removeTab` calls. Before, all parent apps would `removeTab` before all children would. This timing would allow the children to attach to its neighbors on TabGroup disbanding. 
- Fix non-showing tabs on restore.
- Lint.

### DIFF
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -326,7 +326,7 @@ export class DesktopTabGroup implements DesktopEntity {
             }
         });
 
-        if (!activeTabId) {
+        if (activeTabId) {
             // Set the desired tab as active
             await this.switchTab(activeTab);
         } else {
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -326,15 +326,7 @@ export class DesktopTabGroup implements DesktopEntity {
             }
         });
 
-        if (activeTabId) {
-            // Set the desired tab as active
-            await this.switchTab(activeTab);
-        } else {
-            // Need to re-send tab-activated event to ensure tab is active within tabstrip
-            // TODO: See if this can be avoided
-            const event: TabActivatedEvent = {tabstripIdentity: this.identity, identity: activeTab.identity, type: 'tab-activated'};
-            this._window.sendEvent(event);
-        }
+        await this.switchTab(activeTab);
     }
 
     public async swapTab(tabToRemove: DesktopWindow, tabToAdd: DesktopWindow): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/provider/model/DesktopTabGroup.ts
```diff
@@ -323,7 +323,7 @@ export class DesktopTabGroup implements DesktopEntity {
             }
         });
 
-        if (!activeTabId) {
+        if (activeTabId) {
```
ryanpetersonOF: Give this a test, but I think its safe to remove the if statement altogether, including the send event stuff in the else.  You can just make it `await this.switchTab(activeTab);`

src/provider/workspaces/restore.ts
```diff
@@ -189,12 +189,16 @@ const createWorkspacePlaceholders = async(workspace: Workspace): Promise<void> =
...
-                await mainWindowModel.setSnapGroup(new DesktopSnapGroup());
-            }
 
             // Need to check its child windows here, if confirmed.
+            // Also calls removeTab and setSnapGroup on open child windows.
+            // These functions need to be in this order, otherwise child windows may re-attach (if removeTab is called on an application that it's tabbed to,
+            // leaving the child window as a remainingTab).
             await childWindowPlaceholderCheckRunningApp(app, tabbedWindows, tabbedPlaceholdersToWindows, openWindows);
+
+            if (mainWindowModel && mainWindowModel.snapGroup.length > 1) {
```
pjbroadbent: To be honest, not entirely sure what the context is here - but this likely wants to be `mainWindowModel.snapGroup.entities.length`..?

The former checks windows, meaning that a single tabgroup would match this check - and a window would be detached from the rest of it's tabgroup. The latter checks entities, meaning only "proper" docking of windows/tabgroups will trigger this.

src/provider/model/DesktopTabGroup.ts
MichaelMCoates: Cool, sounds good, will do.

src/provider/workspaces/restore.ts
bryangaleOF: We have snapGroup.isNonTrival() now 🙂

MichaelMCoates: So the idea here is that we want to make sure that the window isn't grouped to anything at all before it begins the rest of the restore cycle. If it is grouped, we'll end up dragging the rest of its group along with it wherever we place it, which can lead to really strange looking layouts. That's why we want to separate it. What do you think? This might not be the right approach

pjbroadbent: Makes sense. My only concern was that changing the snap group of just a single window could end up detaching the tabstrip (if that window was tabbed) - but then we remove this window from any tabgroup right above this line.

PR #291, opened 2019-02-21
[SERVICE-360] Possible to disjoint tabs from tabstrips with quick mouse movements during addTab

tomrobertsOF:
Root cause of issue was found to be using stale position from currentState. This PR adds some mitigations including 
- slight delay in addTabInternal to allow bounds-changing events to be processed in the middle
- Added readonly public boolean to DesktopWindow which tracks if the window is moving (set in bounds-changing and unset in bounds-changed).
- Added logic to addTabInternal which will detect if the target window is moving and delay most of the adding operation until the movement has stopped (i.e. the onCommit signal is fired).

### DIFF
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -593,11 +593,29 @@ export class DesktopTabGroup implements DesktopEntity {
             const halfSize: Point = {x: tabState.halfSize.x, y: tabState.halfSize.y - (this._config.height / 2)};
             await tab.applyProperties({center, halfSize, frame: false});
         } else {
+            // Delay to allow other async operations to jump ahead in the queue. Specifically, any pending boundsChanging events
+            // should be processed before continuing.
+            await new Promise(res => setTimeout(res, 10));
+            // If the target window/group is in the process of being moved, we delay the rest of the operation until we receive
+            // a bounds changed and update the currentState. This should serve as a fix/mitigation for SERVICE-360.
+            if (this._window.moveInProgress) {
+                console.log('a');
+                await new Promise(async res => {
+                    const slot = this._window.onCommit.add(async (win, type) => {
+                        console.log('b');
+                        slot.remove();
+                        res();
+                    });
+                    console.log('c');
+                });
+            }
             const existingTabState: EntityState = this._activeTab && this._activeTab.currentState || this._tabs[0].currentState;
             const {center, halfSize} = existingTabState;
-
+            
             // Align tab with existing tab
+            console.log('d');
             await tab.applyProperties({center, halfSize, frame: false});
+            
         }
 
         await tab.setTabGroup(this);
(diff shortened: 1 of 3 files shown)


src/provider/model/DesktopWindow.ts
```diff
@@ -293,18 +294,19 @@ export class DesktopWindow implements DesktopEntity {
...
     private _pendingActions: Promise<void>[];
     private _actionTags: WeakMap<Promise<void>, string>;
-
+    
     // Tracks event listeners registered on the fin window for easier clean-up.
     private _registeredListeners: Map<OpenFinWindowEvent, (event: fin.OpenFinWindowEventMap[OpenFinWindowEvent]) => void> = new Map();
-
+    
+    private _moveInProgress = false;
     private _userInitiatedBoundsChange = false;
```
pjbroadbent: Do we need more state/flags here, or would using `_userInitiatedBoundsChange` produce the same behaviour?

Questionable if we'd want to enable this work-around on non-user-initiated movements, timing would be different - not clear if we'd get the same problem and if the same solution would work..?

tomrobertsOF: I did it this way intentionally because one of the ways I was reliably replicating it in testing was calling `tabGroup._window.moveBy(100,100)` in the console immediately after a `tabGroup.addTab(...)` call without any awaits.

I'd say if we can handle both cases with the same fix than we should.

tomrobertsOF: And for the avoidance of doubt, this does resolve the programmatic one too.

src/provider/model/DesktopTabGroup.ts
```diff
@@ -596,10 +596,24 @@ export class DesktopTabGroup implements DesktopEntity {
...
+                        slot.remove();
+                        res();
+                    });
+                });
+            }
             const existingTabState: EntityState = this._activeTab && this._activeTab.currentState || this._tabs[0].currentState;
             const {center, halfSize} = existingTabState;
 
             // Align tab with existing tab
+            console.log('d');
```
ryanpetersonOF: Can you remove this please?

tomrobertsOF: Already done.

pjbroadbent:
@tomrobertsOF two tests failed in CI. May not be an issue with PR, but can you re-kick and get a clean run?

Will merge as soon as CI passes.

src/provider/model/DesktopTabGroup.ts
```diff
@@ -596,6 +596,19 @@ export class DesktopTabGroup implements DesktopEntity {
...
             await tab.applyProperties({center, halfSize, frame: false});
         } else {
+            // Delay to allow other async operations to jump ahead in the queue. Specifically, any pending boundsChanging events
+            // should be processed before continuing.
+            await new Promise(res => setTimeout(res, 10));
+            // If the target window/group is in the process of being moved, we delay the rest of the operation until we receive
+            // a bounds changed and update the currentState. This should serve as a fix/mitigation for SERVICE-360.
+            if (this._window.moveInProgress) {
+                await new Promise(async res => {
+                    const slot = this._window.onCommit.add(async (win, type) => {
```
MichaelMCoates: Is it possible for the `bounds-changed` event to come in before this `onCommit` is set? Could that make this hang indefinitely?

MichaelMCoates:
CI Passed. Merging.

PR #292, opened 2019-02-21
[SERVICE-354] Calling 'createTabGroup' from a tabbed window causes other windows in the tab group to be ejected to the wrong positions (Partial fix)

bryangaleOF:
This fixes an issue encountered on a similar path to SERVICE-354.

To reproduce the issue, create a tab group of two tabs, and then snap to an untabbed window. Then, call createTabGroupWithTabs for one of the tabbed windows, and the untabbed window. This would result in a tabstrip that was offset from any windowed.

This also adds an await to setSnapGroup in the edited function, out of caution.


bryangaleOF:
Jenkins started at 9dd4460

PR #293, opened 2019-02-21
Add typedoc theme as a submodule & use it in typedoc

aziz512:
Added the [redesigned typedoc template](https://github.com/openfin/typedoc-template) as a submodule and made typedoc use it when generating API docs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "typedoc-template"]
+	path = typedoc-template
+	url = https://github.com/openfin/typedoc-template
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "test": "npm run test:unit && npm run test:int",
     "check": "gts check && cd test && gts check",
     "clean": "gts clean && rimraf gen",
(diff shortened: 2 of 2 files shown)


pjbroadbent:
The default theme had expand/collapse behaviours on overloading functions (e.g. [tabbing.addEventListener](https://cdn.openfin.co/services/openfin/layouts/0.9.6/docs/modules/tabbing.html#addeventlistener)) which don't seem to work in this branch. Is there any way to restore those? The same method in the new docs is now rather long, and not super-readable.

Other than that, looking good. Definitely more consistent with runtime docs. We may move the sub-module around at some point - as we may end up adding a few submodules of our own once we start working on a second service - but can figure that out later.

aziz512:
@pjbroadbent I made some commits to the openfin/typedoc-template. If you pull the changes in the submodule expand/collapse should work. Let me know whether this works or if you find any other issues.

pjbroadbent:
@aziz512 Thanks, can you update the submodule in this branch to point at the new `typedoc-template` commit?

All works if I update the submodule locally, but just checking out the PR branch doesn't have the expand/collapse fix.

aziz512:
@pjbroadbent Cool, I updated the submodule and it should be all good now.

PR #294, opened 2019-02-21
[N/A] Harden API

ryanpetersonOF:
Hardens the client API and standardizes error messages a bit.  

Adds the following API:
`tabSelfTo` and `tabToSelf` and `tabWindowToWindow` (in place of addTabs)

Changes: 
* Can no longer "tab to yourself"
* Can no longer pass duplicate identities into createTabs, otherwise youll end up in a bad state.
* Adds in types for client API payloads
* Added ErrorMessage Enum
* Test bed updated to account for new APIs


src/client/tabbing.ts
```diff
@@ -269,46 +267,48 @@ export async function setTabstrip(config: ApplicationUIConfig): Promise<void> {
...
  *
- * @param windows Array of windows which will be added to the new tab group.
- * @throws `Error`: If no windows is not an array or less than 2 windows were provided.
+ * @param identities Array of window {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities} which will be added to the
+ * new tab group.
+ * @param activeTab The {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity} of the window to set as the active tab in
+ * the group.  If not provided, the first tab in the tab group will be set as the active tab.
+ * @throws `Error`: If one of the provided {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities} is not valid.
+ * @throws `Error`: If duplicate {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities} are provided.
+ * @throws `Error`: If no windows is not an array or less than 2 windows identities were provided.
```
pjbroadbent: Check wording

src/client/tabbing.ts
```diff
@@ -269,46 +267,48 @@ export async function setTabstrip(config: ApplicationUIConfig): Promise<void> {
...
  * tabbing.createTabGroup([{uuid: "App1", name: "App1"}, {uuid: "App2", name: "App2"}, {uuid: "App3", name: "App3"}]);
  * ```
  *
- * @param windows Array of windows which will be added to the new tab group.
- * @throws `Error`: If no windows is not an array or less than 2 windows were provided.
+ * @param identities Array of window {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities} which will be added to the
+ * new tab group.
+ * @param activeTab The {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity} of the window to set as the active tab in
+ * the group.  If not provided, the first tab in the tab group will be set as the active tab.
+ * @throws `Error`: If one of the provided {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities} is not valid.
```
pjbroadbent: This is just the client-side definition of "valid" in this case? Provider will still accept non-tabbed or non-window identities?

Probably makes sense to not be particularly strict in this case, due to the complexity of trying to figure out which of the passed identities was causing the problem within the calling application. Just so long as 2+ of the identities reflect tab-able windows.

src/client/tabbing.ts
```diff
@@ -269,46 +267,48 @@ export async function setTabstrip(config: ApplicationUIConfig): Promise<void> {
...
- *
- * // Tab App2 to App1.
- * tabbing.addTab({uuid: 'App1', name: 'App1'}. {uuid: 'App2', name: 'App2'});
+ * // Tab App1 to App2
+ * tabbing.tabWindowToWindow({uuid: 'App1', name: 'App1'}, {uuid: 'App2', name: 'App2'});
  * ```
  *
  * @param targetWindow The identity of the window to create a tab group on.
- * @param windowToAdd The identity of the window to add to the tab group.  If no `Identity` is provided as an argument the current window context will be used.
+ * @param windowToAdd The identity of the window to add to the tab group.
```
pjbroadbent: Switch ordering of these, to match the order they appear in the function signature.

src/client/tabbing.ts
```diff
@@ -318,7 +318,7 @@ export async function addTab(targetWindow: Identity, windowToAdd: Identity = get
  * ```ts
  * import {tabbing} from 'openfin-layouts';
  *
- * // Remove the current context from its tab group.
+ * // Remove the window from its tab group.  This does not close the window.
  * tabbing.removeTab();
```
pjbroadbent: (referencing the `@throws` below)

Does this function still throw if the window isn't tabbed? May want to make consistent with `undock` API's - if the window ends up in the right state (i.e. untabbed) then that counts as a 'resolve'...?

src/client/tabbing.ts
```diff
@@ -269,46 +267,48 @@ export async function setTabstrip(config: ApplicationUIConfig): Promise<void> {
...
- * @param windowToAdd The identity of the window to add to the tab group.  If no `Identity` is provided as an argument the current window context will be used.
+ * @param windowToAdd The identity of the window to add to the tab group.
  * @throws `Error`: If the {@link ApplicationUIConfig| App Config} does not match between the target and window to add.
  * @throws `Error`: If the `targetWindow` is not a valid {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity}.
  * @throws `Error`: If the `windowToAdd` is not a valid {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity}.
- * @throws `Error`: If `identity` is not an existing tab in a tabstrip.
  */
-export async function addTab(targetWindow: Identity, windowToAdd: Identity = getId()): Promise<void> {
+export async function tabWindowToWindow(windowToAdd: Identity, targetWindow: Identity): Promise<void> {
     return tryServiceDispatch<AddTabPayload, void>(TabAPI.ADDTAB, {targetWindow: parseIdentity(targetWindow), windowToAdd: parseIdentity(windowToAdd)});
```
pjbroadbent: Should we rename the constant here? To match the new API name.

Last chance to make any breaking changes... :-P

src/demo/testbed/TabbingUI.ts
```diff
@@ -99,14 +100,24 @@ export class TabbingUI {
...
+        this._tabSelfTo.onSelect = (identity: WindowIdentity) => {
+            const promise: Promise<void> = tabbing.tabSelfTo(identity);
+
+            this._log.addApiCall(promise, tabbing.tabSelfTo, identity);
+        };
+
         this._addToGroup = new Dropdown(elements.addTab, elements.addTabDropdown);
-        this._addToGroup.header = 'Tabbed Windows';
-        this._addToGroup.placeholder = 'No tabbed windows<br />(Use \'Create Tab Group\')';
-        this._addToGroup.dataProvider = this.getTabbedWindows.bind(this);
```
pjbroadbent: Is `getTabbedWindows` no longer used now?

Could remove the filtering, just have all dropdowns display all windows?..

src/provider/APIHandler.ts
```diff
@@ -2,11 +2,11 @@ import {Identity} from 'hadouken-js-adapter';
...
 import {ChannelProvider} from 'hadouken-js-adapter/out/types/src/api/interappbus/channel/provider';
 
-import {RegisterAPI, SERVICE_CHANNEL, SnapAndDockAPI, TabAPI, WorkspaceAPI} from '../client/internal';
+import {AddTabPayload, CreateTabGroupPayload, RegisterAPI, SERVICE_CHANNEL, SetTabstripPayload, SnapAndDockAPI, TabAPI, UpdateTabPropertiesPayload, WorkspaceAPI} from '../client/internal';
 import {ApplicationUIConfig, TabProperties} from '../client/tabbing';
 
-import {MessageMap} from './APIMessages';
-import {ConfigStore} from './main';
+import {ErrorMessage, MessageMap} from './APIMessages';
+import {ConfigStore, tabService} from './main';
```
pjbroadbent: Using these exports is "deprecated" - they're currently only used in workspaces. Can you replace any uses of this `tabService` import with `this._tabService`?

src/provider/APIHandler.ts
```diff
@@ -192,19 +199,22 @@ export class APIHandler {
         if (tabGroup) {
             return tabGroup.removeTab(ejectedTab!);
         } else if (!ejectedTab) {
-            throw new Error(`No tab with ID ${tab ? `${tab.uuid}/${tab.name}` : tab}`);
+            throw new Error(ErrorMessage.NOWINDOW);
         } else {
-            throw new Error(`Tab with ID ${ejectedTab.id} doesn't belong to a tab group`);
+            throw new Error(ErrorMessage.NOTABGROUP);
```
pjbroadbent: Can we put the parameters back in? Going to make debugging way harder on the application side if we remove any useful information from the error messages.

Standardising the error messages is good, but we should keep the parameter info in there. Maybe some kind of basic templating stuff?

Definitely want to make sure we keep error messages both descriptive and useful.

src/provider/APIMessages.ts
```diff
@@ -24,3 +24,9 @@ export type MessageMap = {
 };
 
 export type EventMap = TabstripEventMap|TabbingEventMap|WorkspacesEventMap|SnapAndDockEventMap;
+
+
+export enum ErrorMessage {
+    NOWINDOW = 'Cannot find Window.  It may be deregistered.',
+    NOTABGROUP = 'No tab group found for the window.'
```
pjbroadbent: `CONST_CASE` for these? As in `NO_TAB_GROUP`, etc?

src/provider/tabbing/TabService.ts
```diff
@@ -81,6 +81,13 @@ export class TabService {
             throw new Error('Must provide at least 2 Tab Identifiers');
         }
 
+        // Checks duplicate entries in the provided array.
+        tabIdentities.forEach(id => {
+            if (tabIdentities.filter(filterId => filterId.name === id.name && filterId.uuid === filterId.uuid).length > 1) {
+                throw new Error('Duplicate identity found.  Provided identites must be unique.');
```
pjbroadbent: State which identity was duplicated.

res/demo/testbed/index.html
```diff
@@ -29,10 +29,16 @@ <h5 class="card-title">Tabbing</h5>
                     <h6 id="tab-status" class="card-subtitle mb-2 text-muted">Not tabbed</h6>
                         <div class="btn-group">
                             <button id="addTab" class="btn btn-primary border-dark dropdown-toggle mb-2" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                                Add to Tab Group
+                                Tab To Self
                                 <div id="addTabDropdown" class="dropdown-menu" aria-labelledby="addTab"></div>
```
bryangaleOF: Worth making `id` and `aria-labelledby` consistent?

src/client/tabbing.ts
```diff
@@ -269,46 +267,48 @@ export async function setTabstrip(config: ApplicationUIConfig): Promise<void> {
 }
 
 /**
- * Given a set of windows, will create a tab group construct and UI around them.  The bounds and positioning of the first (applicable) window in the set will be
- *
- * used as the seed for the tab UI properties.
+ * Creates a tabgroup with the provided windows.  The bounds and positioning of the first (applicable) window in the set will be used as the seed for the tab UI
```
bryangaleOF: I'm unsure what 'applicable' means here. And is there a better wording than 'seed'? Makes me think of random number generation?

src/client/tabbing.ts
```diff
@@ -464,7 +464,7 @@ export async function restoreTabGroup(identity: Identity = getId()): Promise<voi
 }
 
 /**
- * Updates a tab's Properties on the Tab strip.  This includes the tabs title and icon.
+ * Updates a tabs properties. Properties for a tab include its title and icon when in a tab group.
```
bryangaleOF: The apostrophe was correct before? (Or 'Updates the properties of a tab')

src/demo/testbed/View.ts
```diff
@@ -5,6 +5,7 @@ export interface Elements {
     undockGroup: HTMLButtonElement;
     undockWindow: HTMLButtonElement;
     addTab: HTMLButtonElement;
```
bryangaleOF: Consistency with new naming

src/provider/model/DesktopTabGroup.ts
```diff
@@ -468,6 +468,9 @@ export class DesktopTabGroup implements DesktopEntity {
 
             await tab.applyProperties({hidden: false, state: tabState});
 
+            // Fixes flicker on tab switch caused by next window and previous window not showing or hiding before resolving its promise.
+            await new Promise(res => setTimeout(res, 75));
```
bryangaleOF: Why 75? When we're doing this kind of thing elsewhere we're using 10...

Maybe something we could pull out into a constant? INFINTESIMAL_DELAY, LONG_DELAY, etc.? Would also keep track of these kind of solutions that I think ultimately we want to handle better.

bryangaleOF:
A concern I want to raise - we don't have any particularly well defined behaviour when it comes to createTabGroup, and how that interacts with tabs that are already in snap groups, and what snap groups any ejected tab will be in. Right now, we can get weird situations where two overlapping tabgroups will be together in a snap group, and presumably we want to change that. But then that would be changing behaviour of a public API (even if undocumented behaviour)

src/client/tabbing.ts
ryanpetersonOF: I dont know what happened here... fixed!

src/client/tabbing.ts
ryanpetersonOF: The client side definition is any valid Identities.  The provider does filter out any invalid identities and continues on.  There are no checks in place to see if the windows are tabbed or not.

I think for the client side a hard fail if any are invalid makes more sense.  The fail comes from `parseIdentity` which we have on pretty much all of the methods.

src/client/tabbing.ts
ryanpetersonOF: Fixed.  I thought you might pick up on this :P

src/client/tabbing.ts
ryanpetersonOF: Agreed. Fixed

src/client/tabbing.ts
ryanpetersonOF: Do you have any recommendations?  I considered this change during the dev but concluded that we are still "adding a tab" to something.  Its used between all 3 add tab ops (tabToSelf, tabSelfTo, tabWindowToWindow)

src/demo/testbed/TabbingUI.ts
ryanpetersonOF: I dont think its used.  The dataproviders should be giving all windows now in my use cases.

src/provider/APIHandler.ts
ryanpetersonOF: Good catch.  I have used this._tabService instead

src/provider/APIMessages.ts
ryanpetersonOF: Fixed

src/client/tabbing.ts
tomrobertsOF: maybe "initial state" instead of seed

src/provider/APIHandler.ts
ryanpetersonOF: Fixed.  Create templated error creation

src/provider/tabbing/TabService.ts
ryanpetersonOF: Fixed.  Added which identity was duped

res/demo/testbed/index.html
ryanpetersonOF: Im not bothered either way, but the entire file appears to be setup this way.  Worth mentioning with @pjbroadbent

src/client/tabbing.ts
ryanpetersonOF: Yeah a bit strange.  Ive reworded it

src/client/tabbing.ts
ryanpetersonOF: Oops! fixed

src/demo/testbed/View.ts
ryanpetersonOF: fixed

src/provider/model/DesktopTabGroup.ts
ryanpetersonOF: Anything under 50 still produced a flicker.  I gave it some extra room on top.

res/demo/testbed/index.html
pjbroadbent: Not sure I follow, I think Bryan's referring to how the dropdown is (was) named after the API function. The API was renamed (addTab -> tabToSelf), but the `id` and `aria` fields are still named "addTab".

ryanpetersonOF:
@bryangaleOF What do you propose?

res/demo/testbed/index.html
ryanpetersonOF: Well in that case it is fixed...

src/client/tabbing.ts
pjbroadbent: I'd go with `TAB_WINDOW_TO_WINDOW`, assuming that all the other constants match the API name? But only minor, can leave if that's a bit too word-y.

If changing, also update on the provider-side too (in `APIHandler` only, not down into `TabService` etc) and payload names, etc.

bryangaleOF:
@ryanpetersonOF Not proposing anything concrete. But this case in particular raises a question that we'll probably need to answer down the line of how ok we are changing (undocumented) behaviour of API functions

src/provider/APIMessages.ts
```diff
@@ -24,3 +25,42 @@ export type MessageMap = {
...
+     */
+    NO_TAB_GROUP,
+    /**
+     * Unexpected Error
+     */
+    UNEXPECTED
+}
+
+type ErrorMessageArgs = {
+    [ErrorType.NO_WINDOW]: WindowIdentity;[ErrorType.NO_TAB_GROUP]: WindowIdentity;[ErrorType.UNEXPECTED]: {action: string, error: string}
```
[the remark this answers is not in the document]
ryanpetersonOF: This was the linter

src/provider/APIMessages.ts
```diff
@@ -24,3 +25,42 @@ export type MessageMap = {
...
+};
+
+/**
+ * Generates a templated error message.  Does not throw an error.
+ */
+export function createError<T extends keyof ErrorMessageArgs>(msg: T, args: ErrorMessageArgs[T]): string {
```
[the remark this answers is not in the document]
ryanpetersonOF: Ok fixed

src/client/tabbing.ts
ryanpetersonOF: Ok renamed to TAB_WINDOW_TO_WINDOW

src/client/tabbing.ts
```diff
@@ -269,56 +267,57 @@ export async function setTabstrip(config: ApplicationUIConfig): Promise<void> {
...
  * tabbing.createTabGroup([{uuid: "App1", name: "App1"}, {uuid: "App2", name: "App2"}, {uuid: "App3", name: "App3"}]);
  * ```
  *
- * @param windows Array of windows which will be added to the new tab group.
- * @throws `Error`: If no windows is not an array or less than 2 windows were provided.
+ * @param identities Array of window {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities} which will be added to the
+ * new tab group.
+ * @param activeTab The {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity} of the window to set as the active tab in
+ * the group.  If not provided, the first tab in the tab group will be set as the active tab.
+ * @throws `Error`: If one of the provided {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities} is not valid.
```
MichaelMCoates: *Indentitites > Identities

ryanpetersonOF: Where?

src/client/tabbing.ts
```diff
@@ -269,56 +267,57 @@ export async function setTabstrip(config: ApplicationUIConfig): Promise<void> {
...
+ * @param windowToAdd The identity of the window to add to the tab group.
  * @param targetWindow The identity of the window to create a tab group on.
- * @param windowToAdd The identity of the window to add to the tab group.  If no `Identity` is provided as an argument the current window context will be used.
  * @throws `Error`: If the {@link ApplicationUIConfig| App Config} does not match between the target and window to add.
  * @throws `Error`: If the `targetWindow` is not a valid {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity}.
  * @throws `Error`: If the `windowToAdd` is not a valid {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Identity}.
- * @throws `Error`: If `identity` is not an existing tab in a tabstrip.
  */
-export async function addTab(targetWindow: Identity, windowToAdd: Identity = getId()): Promise<void> {
+export async function tabWindowToWindow(windowToAdd: Identity, targetWindow: Identity): Promise<void> {
```
MichaelMCoates: Looks like `targetWindow` and `windowToAdd` was reversed here. Was that intentional?

ryanpetersonOF: Yes to match the new function name.  "tab window TO window" - target window being second

src/client/tabbing.ts
MichaelMCoates: ``Error`: If one of the provided {@link https://developer.openfin.co/docs/javascript/stable/global.html#Identity | Indentities}`

Also in the other comments here

test/client/TabApi.unittest.ts
```diff
@@ -21,27 +21,27 @@ beforeEach(async () => {
...
     describe('When calling addTab', () => {
         it('Calling with invalid identity rejects with error message', async () => {
-            const promise = tabbing.addTab(null!, null!);
+            const promise = tabbing.tabWindowToWindow(null!, null!);
             await expect(promise).rejects.toThrowError(ErrorMsgs.INVALID_IDENTITY);
         });
 
         it('Calling with invalid uuid rejects with error message', async () => {
-            const promise = tabbing.addTab({uuid: null!, name: 'somename'}, null!);
+            const promise = tabbing.tabWindowToWindow({uuid: null!, name: 'somename'}, null!);
```
MichaelMCoates: Looks like these args weren't reversed with the `tabWindowToWindow` arg reversal. Should be fine, but wanted to note in case you missed it.

ryanpetersonOF: It will fail either way

src/client/tabbing.ts
ryanpetersonOF: Ah, a little different than `Indentitites` :P

src/client/tabbing.ts
pjbroadbent: Yeah, we thought this was a more understandable ordering. Order of parameters matches the name of the function.

i.e: `tabXToY(x: Identity, y: Identity)`

src/client/tabbing.ts
ryanpetersonOF: Fixed

PR #295, opened 2019-02-21
[SERVICE-358] Possible to close a tabstrip independent of its tabs

bryangaleOF:
DesktopTabGroup listens for teardown of tabstrip window and closes all tabs as needed

Passed Jenkins at d5060e5

### DIFF
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -681,8 +681,6 @@ export class DesktopTabGroup implements DesktopEntity {
     private async onTabGroupTeardown(): Promise<void> {
         if (!this._closingOnTabRemoval) {
             return this.removeAllTabs(true);
-        } else {
-            return Promise.resolve();
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/provider/model/DesktopTabGroup.ts
```diff
@@ -660,6 +665,14 @@ export class DesktopTabGroup implements DesktopEntity {
         }
     }
 
+    private async onTabGroupTeardown(): Promise<void> {
+        if (!this._closingOnTabRemoval) {
```
pjbroadbent: Worth checking if `!this._window.isReady` would work here. Seems like this flag is only set when the window is (about to) close, so should be equivalent to the tabstrip being closed manually.

bryangaleOF: Sorry, doesn't work. And had to stop and think for a bit if it should be ` isReady` or `!isReady` here, so that makes me more comfortable with the explicitness.

pjbroadbent: Oops, yeah - `isReady`. OK, keep it explicit then.

PR #296, opened 2019-02-22
[SERVICE-372] Remove display scaling runtime flags, message any resulting issues to users

tomrobertsOF:
- removed `high-dpi-support` and `force-device-scale-factor` runtime flags.
- will now display error message and disable the service if scaling is not 100%


tomrobertsOF:
As a note, the html and css for the error window are pulled directly from the renderer crash window.

src/provider/main.ts
```diff
@@ -32,6 +33,19 @@ declare const window: Window&{
 fin.desktop.main(main);
 
 export async function main() {
+    const monitorInfo = await fin.System.getMonitorInfo();
+
+    // Disable the service if display scaling is not 100%
+    if (monitorInfo.deviceScaleFactor !== 1) {
```
pjbroadbent: `deviceScaleFactor` also exists on each monitor object. Do we know how the top-level value is calculated?

And do we know what exactly our requirements are on this - as in, what happens if you have some 100% monitors and some non-100% monitors?

res/provider/errors/error.html
```diff
@@ -0,0 +1,147 @@
...
+				return wnd.close();
+			} catch (err) {
+				console.log(err);
+			}
+		}
+
+		async function getExceptionData() {
+			const wnd = await fin.Window.getCurrent();
+			const { customData: { title, error } } = await wnd.getOptions();
+			const errorTitleEl = document.querySelector('#error-title');
```
bryangaleOF: We're using `Elem` elsewhere as the abbreviation for Element. Worth standardizing?

res/provider/errors/error.html
```diff
@@ -0,0 +1,147 @@
...
+		async function closeWin() {
+			try {
+				const wnd = await fin.Window.getCurrent();
+				return wnd.close();
+			} catch (err) {
+				console.log(err);
+			}
+		}
+
+		async function getExceptionData() {
```
bryangaleOF: Would consider renaming as it's much more than a getter. presentExceptionData?

res/provider/errors/error.html
tomrobertsOF: This file is copied verbatim from the core, so I'd prefer to leave as-is for constancy.

res/provider/errors/error.html
tomrobertsOF: As above.

src/provider/main.ts
tomrobertsOF: Not sure how it's decided. I can see in the core that this just calls a custom electron API function.

Need to test out on a multi-monitor setup to see what possible combinations there are.

tomrobertsOF: So it tested fairly extensively and it seems like as long as the primary monitor has 100% display scaling it works fine (technically it's from the monitor that the layouts service window starts on, but they should always be the same), and conveniently enough, the deviceScaleFactor at the top level is the scale factor of the primary monitor.

To summarise then: the error should show up if and only if the scaling is set to a number that would cause problems. Otherwise things will work as normal.

PR #300, opened 2019-02-23
[SERVICE-357] Prevented tabbing "through" de-registered windows.

pjbroadbent:
The model's `getWindowAt` method will no longer return windows that are covered by a de-registered window

This was achieved by using the `ZIndexer` to perform the actual check (since it also tracks non-registered windows), which required having the `ZIndexer` also log the bounds of a window. No extra events were required to do this, instead the util stores the bounds data that comes from the `bounds-changed` events.

Only `bounds-changed` events are used for the position tracking, meaning that data for any window currently being moved will be stale. In actual usage, any active window is always excluded from the `getWindowAt` check.

### DIFF
--- a/src/provider/model/DesktopModel.ts
+++ b/src/provider/model/DesktopModel.ts
@@ -140,8 +140,22 @@ export class DesktopModel {
     }
 
     public getWindowAt(x: number, y: number, exclude?: WindowIdentity): DesktopWindow|null {
-        const identityAtPoint: WindowIdentity|null = this._zIndexer.getWindowAt(x, y, exclude);
-        return identityAtPoint && this.getWindow(identityAtPoint);
+        const point: Point = {x, y};
+        const excludeId: string|undefined = exclude && this.getId(exclude);
+        const windowsAtPoint: DesktopWindow[] = this._windows.filter((window: DesktopWindow) => {
+            const state: EntityState = window.currentState;
+            return window.isActive && RectUtils.isPointInRect(state.center, state.halfSize, point) && window.id !== excludeId;
+        });
+
+        const topMostWindow: DesktopWindow|null = this._zIndexer.getTopMost(windowsAtPoint);
+        const nonModelWindow: WindowIdentity|null = this._zIndexer.getWindowAt(x, y, exclude);
+
+        if (!topMostWindow || !nonModelWindow || topMostWindow.id === this.getId(nonModelWindow!)) {
+            return topMostWindow;
+        } else {
+            // Model found a window at this point, but it is obscured by a de-registered window
+            return null;
+        }
     }
 
     public getTabGroup(id: string): DesktopTabGroup|null {
--- a/src/provider/model/DesktopModel.ts
+++ b/src/provider/model/DesktopModel.ts
@@ -142,18 +142,19 @@ export class DesktopModel {
     public getWindowAt(x: number, y: number, exclude?: WindowIdentity): DesktopWindow|null {
         const point: Point = {x, y};
         const excludeId: string|undefined = exclude && this.getId(exclude);
-        const windowsAtPoint: DesktopWindow[] = this._windows.filter((window: DesktopWindow) => {
+        const modelWindowsAtPoint: DesktopWindow[] = this._windows.filter((window: DesktopWindow) => {
             const state: EntityState = window.currentState;
             return window.isActive && RectUtils.isPointInRect(state.center, state.halfSize, point) && window.id !== excludeId;
         });
 
-        const topMostWindow: DesktopWindow|null = this._zIndexer.getTopMost(windowsAtPoint);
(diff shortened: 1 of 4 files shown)


tomrobertsOF:
One small issue from trying this out, it looks like you are using the unadjusted bounds for non-registered windows. This makes it so you can't tab in the deregistered window's shadow even though it looks like you should be able to.

src/provider/model/DesktopModel.ts
```diff
@@ -147,7 +147,15 @@ export class DesktopModel {
             return window.isActive && RectUtils.isPointInRect(state.center, state.halfSize, point) && window.id !== excludeId;
         });
 
-        return this._zIndexer.getTopMost(windowsAtPoint);
+        const topMostWindow: DesktopWindow|null = this._zIndexer.getTopMost(windowsAtPoint);
+        const nonModelWindow: WindowIdentity|null = this._zIndexer.getWindowAt(x, y, exclude);
+
+        if (!topMostWindow || !nonModelWindow || topMostWindow.id === this.getId(nonModelWindow!)) {
```
bryangaleOF: It's not obvious that `topMostWindow.id === this.getId(nonModelWindow!)` is the check we want. If this is true, we will return a non-model window, right?

test/demo/workspaces/formerTabGroupRestoreResize.test.ts
```diff
@@ -1,31 +1,37 @@
...
 const combinedProgrammaticApps = registeredProgrammaticApp.concat(deregisteredProgrammaticParentandChild);
 
 const registeredManifestApp = createAppsArray(1, 0, {manifest: true, url: 'http://localhost:1337/test/saveRestoreTestingApp.html?deregistered=false'});
 const deregisteredManifestParentandChild =
-    createAppsArray(1, 1, {manifest: true, url: 'http://localhost:1337/test/saveRestoreTestingApp.html?deregistered=true'});
+    createAppsArray(1, 2, {manifest: true, url: 'http://localhost:1337/test/saveRestoreTestingApp.html?deregistered=true'});
 const combinedManifestApps = registeredManifestApp.concat(deregisteredManifestParentandChild);
 
-const windowGrouping = [[0, 2]];
+const windowGrouping = [[3, 1], [0, 2]];
```
bryangaleOF: Feel like I need this change explained to me

test/provider/tabOnDragOver.test.ts
```diff
@@ -511,3 +511,41 @@ test('Drag window into tabgroup then tearout - window should be maximizable if-a
...
+    });
+    wins.push(win3);
+
+    await delay(500);
+    await dragWindowTo(wins[1], 250, 120);
+    await delay(500);
+    await Promise.all(wins.map(win => assertNotTabbed(win, t)));
+});
+
+test('Cannot tab to an obscures window that is not registered to the service', async t => {
```
bryangaleOF: 'Cannot tab to a window that is obscured by a window not registered to the service'?

test/demo/workspaces/formerTabGroupRestoreResize.test.ts
pjbroadbent: The setup for this test previously created three windows. Because of how the application utils work, two of these windows (one registered, one de-registered) end up being created in the same place - with the de-registered window (somehow) always on top.

I added an extra child window (which will always be to the right, not overlapping any other windows), and made the first step of the test to "tab" the de-registered window with this out-of-the-way window.

The windows won't actually tab (since one is de-registered), but it gets the window out of the way, so that the test can actually tab the windows that need to be tabbed. A better solution would have been to make the two windows not be created at the same position, but that isn't possible with the current utils.

src/provider/model/DesktopModel.ts
pjbroadbent: Hmm, name isn't clear - should really be more like `potentiallyNonModelWindow`. In common usage, both `topMostWindow` and `nonModelWindow` will be the same window.

We're basically asking the model what window is at (x,y) and then checking that the ZIndexer (which also tracks de-registered windows) doesn't think that there is another window at that position.

Will take another look, see if there's a way to clarify the logic here.

pjbroadbent:
@tomrobertsOF Started making the bounds change, but that is deceptively complicated in current state of things. The `ZIndexer` doesn't currently track `framed` or any other window options, and whilst we can fetch it when we first register a window (registration is now async, due to requiring the bounds of a window), there's no clean way to ensure this remains correct when tabbing/untabbing or via application changes.

Suggest that we leave as-is for now, and figure out ways to integrate `ZIndexer` into `DesktopModel` some point down the line.

tomrobertsOF:
👍 assuming we've made a jira to prevent forgetting

pjbroadbent:
Tracked as SERVICE-374

PR #302, opened 2019-02-23
[SERVICE-373] Tab properties now initialised with window title

pjbroadbent:
Using `getInfo()` to fetch document title from each window. This isn't always immediately available, so service will re-request the window title when tabbing, if the title isn't already known.

It seems the API returns the URL of the page in cases where there is no `<title>` tag in the DOM. The service will detect these cases, and use the window name instead (as per earlier behaviour).

### DIFF
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -149,6 +149,15 @@ export class DesktopWindow implements DesktopEntity {
                     }
                 };
 
+                // Get window title
+                const {title, url} = info;
+                let windowTitle = title;
+                if (title && (title === url || title === url.substr(url.indexOf('://') + 3))) {
+                    // If there is no <title> tag on document, getInfo will return the URL (minus protocol) as the title.
+                    // In these cases, the actual title will be the window name - not it's URL
+                    windowTitle = window.identity.name || '';
+                }
+
                 return {
                     center,
                     halfSize,
@@ -157,7 +166,7 @@ export class DesktopWindow implements DesktopEntity {
                     hidden: !results[2],
                     state: options.state!,
                     icon: options.icon || `https://www.google.com/s2/favicons?domain=${options.url}`,
-                    title: info.title!,
+                    title: windowTitle,
                     showTaskbarIcon: options.showTaskbarIcon!,
                     opacity: options.opacity!,
                     alwaysOnTop: options.alwaysOnTop!,
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -119,6 +119,7 @@ export class DesktopWindow implements DesktopEntity {
             .then((results: [fin.WindowOptions, WindowInfo, boolean, fin.WindowBounds]): EntityState => {
                 const options: fin.WindowOptions = results[0];
                 const info: WindowInfo = results[1];
+                const isShowing: boolean = results[2];
                 const bounds: fin.WindowBounds = results[3];
                 const halfSize: Point = {x: bounds.width / 2, y: bounds.height / 2};
                 const center: Point = {x: bounds.left + halfSize.x, y: bounds.top + halfSize.y};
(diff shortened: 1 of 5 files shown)


test/provider/tabOnDragOver.test.ts
```diff
@@ -1,27 +1,47 @@
...
+async function getTabTitle(identity: Identity) {
+    return executeJavascriptOnService(function(this: ProviderWindow, tabIdentity: WindowIdentity) {
+        return new Promise((resolve, reject) => {
+            const tab = this.model.getWindow(tabIdentity)!;
+            const tabGroup = tab.tabGroup!;
+            const tabIndex = tabGroup.tabs.findIndex(t => t.id === tab.id);
+            const tabstripIdentity = tabGroup.identity;
+
+            const tabstrip = fin.desktop.Window.wrap(tabstripIdentity.uuid, tabstripIdentity.name!);
+            tabstrip.executeJavaScript(
```
tomrobertsOF: I know this is only in the tests, but are we confident this will always work properly. I have some vague memories from writing the remote exec utils that there are heavy restrictions on using `executeJavaScript()`

src/provider/model/DesktopWindow.ts
```diff
@@ -117,10 +115,11 @@ export class DesktopWindow implements DesktopEntity {
...
 
     public static async getWindowState(window: Window): Promise<EntityState> {
-        return Promise.all([window.getOptions(), window.isShowing(), window.getBounds()])
-            .then((results: [fin.WindowOptions, boolean, fin.WindowBounds]): EntityState => {
+        return Promise.all([window.getOptions(), window.getInfo(), window.isShowing(), window.getBounds()])
+            .then((results: [fin.WindowOptions, WindowInfo, boolean, fin.WindowBounds]): EntityState => {
                 const options: fin.WindowOptions = results[0];
-                const bounds: fin.WindowBounds = results[2];
+                const info: WindowInfo = results[1];
+                const bounds: fin.WindowBounds = results[3];
```
bryangaleOF: Small 'huh' on reading through this that we extract 0, 1, and 3 but not 2. Worth extracting 2 (isShowing) just for the sake of completeness here?

src/provider/model/DesktopWindow.ts
```diff
@@ -150,15 +149,24 @@ export class DesktopWindow implements DesktopEntity {
                     }
                 };
 
+                // Get window title
+                const {title, url} = info;
+                let windowTitle = title;
+                if (title && (title === url || title === url.substr(url.indexOf('://') + 3))) {
+                    // If there is no <title> tag on document, getInfo will return the URL (minus protocol) as the title.
```
bryangaleOF: If it's 'url minus protocol' why also check for `title === url`? Also, I'd put 'probably' somewhere in the comment, since someone might deliberately put the URL as a title, and we're just making a rough guess

src/provider/tabbing/TabService.ts
```diff
@@ -227,21 +228,38 @@ export class TabService {
...
 
     public updateTabProperties(tab: DesktopWindow, properties: Partial<TabProperties>): void {
         const tabProps: TabProperties = this.getTabProperties(tab);
-        Object.assign(tabProps, properties);
-        localStorage.setItem(tab.id, JSON.stringify(tabProps));
+        const newProps: TabProperties = {...tabProps, ...properties};
 
-        const event: TabPropertiesUpdatedEvent = {identity: tab.identity, properties: tabProps, type: 'tab-properties-updated'};
-        tab.sendEvent(event);
+        if (newProps.icon !== tabProps.icon || newProps.title !== tabProps.title) {
```
bryangaleOF: I don't understand why these are the only two circumstances we actually want to do anything

test/provider/tabOnDragOver.test.ts
pjbroadbent: Yeah, I _believe_ this is always safe. It's wrapped in an `executeJavascriptOnService`, so in the context of the actual `executeJavsScript` call, it's looking at a child window.

I did have to use the v1 API for it, as it seems the v2 call always returns `undefined`, rather than the result of the expression.

src/provider/model/DesktopWindow.ts
pjbroadbent: It's extracted below, but yeah - would be clearer pulled-up here.

test/provider/tabOnDragOver.test.ts
tomrobertsOF: Yea, the V2 thing is what I remember. It's basically completely locked down. If we're using the v1 api anyway, could we not just use nativeWindow?

src/provider/tabbing/TabService.ts
pjbroadbent: It's a "only fire an event if the properties have changed" guard.

There's only two properties, the service checks that something has actually changed before continuing. Could use a `deepEqual` to do the comparison, make it both less "arbitrary" and more robust to property changes/additions?

test/provider/tabOnDragOver.test.ts
pjbroadbent: Tried that but couldn't, and not sure why. May only get nativeWindow if it's the same window object you used to create the window, or something along those lines.

src/provider/model/DesktopWindow.ts
pjbroadbent: Nick dug-up the spec for this field, will update to look for the specific string the core uses.

Details here: https://appoji.jira.com/browse/RUN-3457

pjbroadbent: Turns out this story was incorrectly marked as resolved - not yet been implemented.

Checking against the current and future strings, but just those two specific strings - have removed the 'title === url' case.

Since this is now a more in-depth check, check is wrapped in a guard that will avoid this check in normal usage but still preserve behaviour.

src/provider/tabbing/TabService.ts
pjbroadbent: Updated to use `deepEqual`

test/provider/tabOnDragOver.test.ts
tomrobertsOF: Hmm, another one of those undocumented security improvements. Ok then.

PR #304, opened 2019-02-25
[SERVICE-354] Calling 'createTabGroup' from a tabbed window causes other windows in the tab group to be ejected to the wrong positions

bryangaleOF:
Much of SERVICE-354 was resolved incidentally by other PRs, but this wraps up one remaining known issue, and adds tests around `createTabGroupWithTabs`.

Parts to this PR:
- Addition of `promiseForEach` and `promiseFilter` as analogous to `promiseMap`

- Some changes to assertion naming, so `assertTabbed` and `assertNotTabbed ` could work as expected from naming. There's an implied further change to make `assertGrouped` and `assertTabbed` work the similarly, though this isn't part of this PR.

- Additional assertions that introduce the concept of a 'complete' group, that lets us ask if a list of windows form the complete contents of a group and that there are no other members.

- Additional step in `createTabGroupWithTabs` to place ejected tabs in a new SnapGroup, to prevent an issue where there'd be two fully overlapping tab groups, that are in the same SnapGroup

- Parameterized tests around `createTabGroupWithTabs`

Jenkins in progress at dcd772a.


test/demo/tabbing/createTabGroupFromTabs.test.ts
```diff
@@ -0,0 +1,258 @@
...
+
+        const newTabs = testOptions.newTabGroup.map(index => windows[index]);
+        const ejectedTabs = testOptions.ejectedTabGroup.map(index => windows[index]);
+
+        await setupSnapAndTabGroups(t, testOptions);
+
+        const oldBounds = await getBoundsMap(windows);
+
+        function remoteCreateTabGroupWithTabs(this: ProviderWindow, identities: WindowIdentity[]) {
+            return this.tabService.createTabGroupWithTabs(identities);
```
tomrobertsOF: Is there a reason why this needs to use the internal method instead of just using the client's `tabbing.createTabGroup(...)` method?

bryangaleOF: I'd been considering createTabGroupWithTabs as the object under test, but have revised to start at the API layer

tomrobertsOF: Ok, cool. There's not a huge difference functionally, just a bit more consistent across the tests.

PR #306, opened 2019-02-25
[SERVICE-361] Certain configurations of snapped windows will allow you to snap a new window partially on top of other windows.

bryangaleOF:
Jenkins started at 531d9c4.


src/provider/snapanddock/Projector.ts
```diff
@@ -281,7 +281,7 @@ class BorderProjection implements Range {
             if (other.limit < mid) {
                 this.min = Math.max(this.min, other.limit);
             } else {
-                this.max = Math.min(this.min, other.limit);
+                this.max = Math.min(this.max, other.limit);
```
tomrobertsOF: Great catch! No idea how I missed that in the first place.

PR #307, opened 2019-02-25
[N/A] Sequentially launch applications during restoration to avoid createFromManifest timeout

MichaelMCoates:
*DO NOT MERGE*

When restoring a large workspace that uses many applications created from manifest, the core/RVM don't currently account for that level of load. If application launches hang a bit long, they timeout silently, which breaks restoration. 

By restoring the applications sequentially vs. all at once, we avoid this timeout. Temporary solution until RUN-5040 and RVM-814 are resolved.

### DIFF
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -70,14 +70,7 @@ export const restoreWorkspace = async(payload: Workspace): Promise<Workspace> =>
 
     await createWorkspacePlaceholders(workspace);
 
-    // Have to restore the applications one-by-one in order to prevent createFromManifest calls to the RVM from hanging
-    // Should be resolved by RUN-5040 and RVM-814
-    const restoreAppResponses: WorkspaceApp[] = [];
-    for (let index = 0; index < workspace.apps.length; index++) {
-        const workspaceApp = workspace.apps[index];
-        const restoredWorkspaceApp = await restoreApp(workspaceApp, startupApps);
-        restoreAppResponses.push(restoredWorkspaceApp);
-    }
+    const restoreAppResponses: WorkspaceApp[] = await promiseMap(workspace.apps, (app, idx) => restoreApp(app, idx, startupApps));
 
     // Wait for all apps to startup
     const startupResponses: WorkspaceApp[] = await Promise.all(startupApps);
@@ -247,7 +240,9 @@ const createWorkspacePlaceholders = async(workspace: Workspace): Promise<void> =
     await tabService.createTabGroupsFromWorkspace(workspace.tabGroups);
 };
 
-const restoreApp = async(app: WorkspaceApp, startupApps: Promise<WorkspaceApp>[]): Promise<WorkspaceApp> => {
+const delay = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+const restoreApp = async(app: WorkspaceApp, delayIdx: number, startupApps: Promise<WorkspaceApp>[]): Promise<WorkspaceApp> => {
     // Get rid of childWindows for default response (anything else?)
     const defaultResponse = {...app, childWindows: []};
     try {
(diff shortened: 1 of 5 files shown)


bryangaleOF:
What was the old timeout?

MichaelMCoates:
@bryangaleOF The default timeout is 1 second.

src/provider/workspaces/restore.ts
```diff
@@ -276,7 +283,10 @@ const restoreApp = async(app: WorkspaceApp, startupApps: Promise<WorkspaceApp>[]
                 // If app created by manifest
                 const {manifestUrl} = app;
                 console.log('App has manifestUrl:', app);
+                console.log("BEFORE CREATEFROMMANIFEST");
```
tomrobertsOF: Assuming this is some leftover debug logging. Can you remove.

src/provider/workspaces/restore.ts
```diff
@@ -288,7 +298,9 @@ const restoreApp = async(app: WorkspaceApp, startupApps: Promise<WorkspaceApp>[]
             }
 
             if (ofAppNotRunning) {
+                console.log("BEFORE RUN");
```
tomrobertsOF: As above, debug logging

src/provider/workspaces/restore.ts
```diff
@@ -70,7 +70,14 @@ export const restoreWorkspace = async(payload: Workspace): Promise<Workspace> =>
...
 
-    const apps: WorkspaceApp[] = await promiseMap(workspace.apps, app => restoreApp(app, startupApps));
+    // Have to restore the applications one-by-one in order to prevent createFromManifest calls to the RVM from hanging
+    // Should be resolved by RUN-5040 and RVM-814
+    const restoreAppResponses: WorkspaceApp[] = [];
+    for (let index = 0; index < workspace.apps.length; index++) {
+        const workspaceApp = workspace.apps[index];
+        const restoredWorkspaceApp = await restoreApp(workspaceApp, startupApps);
+        restoreAppResponses.push(restoredWorkspaceApp);
+    }
```
bryangaleOF: We have promiseForEach now that could be used here. But would love something like promiseSequentialMap to keep this as a one-liner.

src/provider/workspaces/restore.ts
bryangaleOF: Linter dislikes double quotes

src/provider/workspaces/restore.ts
bryangaleOF: Linter dislikes double quotes

src/provider/workspaces/restore.ts
tomrobertsOF: A `for..of` block with awaits is basically the same and more in-line with what we're doing elsewhere in the code.

bryangaleOF:
Kicked off Jenkins - 3f798ac merged with aa747c

bryangaleOF:
Having another look over, I've realised this will change failure behaviour. Consider a case where the first of mulitple WorkspaceApps we try to restore is an app that is already running, and has a misbehaving restore handler.

In this case, no further apps will restore until the one minute timeout has elapsed, where before this one misbehaving app would not delay other apps in the current run of restore from restoring.

A potential fix (that would need more thinking about just in case there's problems I've missed), would be to change the line:
`return await clientRestoreAppWithTimeout(app);'
to:
'startupApps.push(clientRestoreAppWithTimeout(app)); return defaultResponse;`

I don't think this is necessarily a blocker to this PR going in, but I think it needs to be thought about and at least flagged with Phil.

MichaelMCoates:
Yeah I had similar thoughts as well. I'll keep messing with it and get back to you, thank you for the suggestions

MichaelMCoates:
Changed from launching the applications in a for loop to using an incrementing delay before `createFromManifest` calls. It'll avoid the error Bryan mentioned above. It isn't perfect, and can still hang if you're restoring an insane amount of manifest applications (25+), but there's an RVM fix incoming that helps this a lot.

MichaelMCoates:
Closing in favor of the timeout fix currently in develop.

PR #308, opened 2019-02-26
[N/A] Moved API/config doc location when deploying to the CDN

pjbroadbent:
Have changed the location the generated docs are deployed to as part of the build process.

Now deployed to `/docs/services/layouts/<version>`, as per other projects. Docs folder is still sub-divided into `api` and `config`.


bryangaleOF:
Kicked off Jenkins at 49c8a5 merged with aa747c

PR #309, opened 2019-02-26
[SERVICE-375] Undock docked windows when maximizing

tomrobertsOF:
This was accomplished by adding a call to `snapGroup.validate()` in the maximize event callback on `DesktopWindow`s. Since we now track the visually apparent bounds of maximised windows, the service will recognise that the group is no longer contiguous and take the appropriate actions (e.g. undocking windows or splitting the group).

I expect that this is still some very small time windows where moving the maximized window could cause issues. This is pretty much unavoidable as we are only able to react to the event, not intercept it mid-flight. I have not been able to reproduce this myself, so I expect the window is small enough to not be a major problem.


tomrobertsOF:
Jenkins all passed with the exception of the one test which is also failing on develop. Can merge once we have a second approval.

PR #310, opened 2019-02-26
[SERVICE-357-fix] Update ZIndexer to track 'active' state of windows 

bryangaleOF:
This resolves a test failure caused by ZIndexer not responding to windows being hidden, and so believing hidden windows to be obscuring the expected active window.

Jenkins succeeded at a9d0349.

### DIFF
--- a/src/provider/model/ZIndexer.ts
+++ b/src/provider/model/ZIndexer.ts
@@ -64,14 +64,6 @@ export class ZIndexer {
     }
 
     /**
-     * Returns the array of indexes.
-     * @returns {ZIndex[]} ZIndex[]
-     */
-    public get indexes(): ReadonlyArray<ZIndex> {
-        return this._stack;
-    }
-
-    /**
      * Takes a list of window-like items and returns the top-most item from the list.
      *
      * NOTE: Implementation will not return any item within the input that does not exist within the ZIndexer util.
@@ -110,27 +102,6 @@ export class ZIndexer {
         });
 
         return entry ? entry.identity : null;
-    }
-
-    /**
-     * Takes a list of window-like items, and sorts them by the z-index of their respective windows.
-     *
-     * NOTE: Implementation will filter-out any windows within the input that do not exist within the ZIndexer util.
-     *
-     * @param items
-     */
-    public sort<T extends Identifiable>(items: T[]): T[] {
-        const ids: string[] = this.getIds(items);
-        const result: T[] = [];
-
-        for (const item of this._stack) {
-            const index: number = ids.indexOf(item.id);
-            if (index >= 0) {
-                result.push(items[index]);
-            }
-        }
-
-        return result;
     }
 
     /**
--- a/src/provider/model/ZIndexer.ts
+++ b/src/provider/model/ZIndexer.ts
@@ -108,7 +108,7 @@ export class ZIndexer {
     /**
      * Updates the windows index in the stack and sorts array.
      * @param identity ID of the window to update (uuid, name)
-     * @param bounds Physical bounds of the window to update, if known. Will perform an async querey if needed and not specified
+     * @param bounds Physical bounds of the window to update, if known. Will perform an async query if needed and not specified
      * @param active The active state of the window to update, if know. Will guess as true if needed and not specified
      * @param timestamp The new timestamp for the stack entry. Will use the current time if not specified
      */
--- a/src/provider/model/ZIndexer.ts
+++ b/src/provider/model/ZIndexer.ts
@@ -75,10 +75,12 @@ export class ZIndexer {
     public getTopMost<T extends Identifiable>(items: T[]): T|null {
         const ids: string[] = this.getIds(items);
 
-        for (const item of this._stack.filter(entry => entry.active)) {
-            const index: number = ids.indexOf(item.id);
-            if (index >= 0) {
-                return items[index];
+        for (const item of this._stack) {
+            if (item.active) {
+                const index: number = ids.indexOf(item.id);
+                if (index >= 0) {
+                    return items[index];
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/provider/model/ZIndexer.ts
```diff
@@ -112,79 +105,60 @@ export class ZIndexer {
...
-        }
-
-        return result;
-    }
-
     /**
      * Updates the windows index in the stack and sorts array.
      * @param identity ID of the window to update (uuid, name)
-     * @param bounds Physical bounds of the window to update, if known
+     * @param bounds Physical bounds of the window to update, if known. Will perform an async query if needed and not specified
```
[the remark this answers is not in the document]
bryangaleOF: Thanks, fixed

src/provider/model/ZIndexer.ts
```diff
@@ -93,7 +86,7 @@ export class ZIndexer {
     }
 
     public getWindowAt(x: number, y: number, exclude?: WindowIdentity): WindowIdentity|null {
-        const entry: ZIndex|undefined = this._stack.find((item: ZIndex) => {
+        const entry: ZIndex|undefined = this._stack.filter(entry => entry.active).find((item: ZIndex) => {
```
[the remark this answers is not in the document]
bryangaleOF: Yeah, debated just doing this check in the find, but felt nicer to have a consistent 'remove inactive' pass done the same way as in getTopMost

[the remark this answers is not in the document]
tomrobertsOF: My main concern is that this will potentially be called every frame during a tabbing operation and so any easy performance gains should be considered.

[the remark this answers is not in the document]
bryangaleOF: Ah k 👍. Changed.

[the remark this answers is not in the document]
tomrobertsOF: Awesome, thanks!

src/provider/model/ZIndexer.ts
```diff
@@ -112,79 +105,60 @@ export class ZIndexer {
...
-
-        return result;
-    }
-
     /**
      * Updates the windows index in the stack and sorts array.
      * @param identity ID of the window to update (uuid, name)
-     * @param bounds Physical bounds of the window to update, if known
+     * @param bounds Physical bounds of the window to update, if known. Will perform an async query if needed and not specified
+     * @param active The active state of the window to update, if know. Will guess as true if needed and not specified
```
pjbroadbent: Typo, `know` -> `known`.

src/provider/model/ZIndexer.ts
```diff
@@ -112,79 +105,60 @@ export class ZIndexer {
...
-
-                if (entry) {
-                    // Update bounds on existing entry
-                    Object.assign(entry.bounds, bounds);
-                } else {
-                    // Continue with creating new entry
-                    this.addToStack({id, identity, timestamp, bounds: this.sanitizeBounds(bounds)});
-                }
+                // Since this required an async operation, entry may now exist within stack, so recursively call update
+                this.update(identity, active, this.sanitizeBounds(bounds), timestamp);
```
pjbroadbent: Seems this is the only place a `timestamp` is passed to `update`.

Is it important that we keep the "initial" timestamp in this case? Would make sense to me to use the current timestamp (as that's the timestamp of that bounds data), and simplify the method a bit.

src/provider/model/ZIndexer.ts
```diff
@@ -227,6 +210,18 @@ export class ZIndexer {
         this._stack.push(entry);
     }
 
+    private sortStack(): void {
+        this._stack.sort((a, b) => {
+            if (a && !b) {
```
pjbroadbent: `a` => `a.active`? etc.

bryangaleOF: Fixed.

src/provider/model/ZIndexer.ts
bryangaleOF: Fixed.

src/provider/model/ZIndexer.ts
bryangaleOF: I'd argue yes. The use of the timestamp, rather than say a counter that gets incremented in each event response, is already a 'this is *probably* correct' thing. Allowing the delay on the getBounds call to influence things makes that 'probably' bigger

bryangaleOF:
@tomrobertsOF Rejecting your previous review, would be good to get your eyes on the new solution

src/provider/model/ZIndexer.ts
```diff
@@ -204,8 +204,10 @@ export class ZIndexer {
         win.addListener('bounds-changed', boundsChanged);
 
         // When a window is hidden or minimized, mark it as inactive
-        win.addListener('hidden', markInactive);
-        win.addListener('minimized', markInactive);
+        // Todo: Renable these to allow tracking of hidden non-Model windows.
```
tomrobertsOF: Can we make a JIRA and reference it here. Don't like random TODO comments hanging around.

src/provider/model/ZIndexer.ts
```diff
@@ -187,8 +187,8 @@ export class ZIndexer {
             win.removeListener('focused', bringToFront);
             win.removeListener('shown', bringToFront);
             win.removeListener('bounds-changed', boundsChanged);
-            win.removeListener('hidden', markInactive);
-            win.removeListener('minimized', markInactive);
+            // win.removeListener('hidden', markInactive);
```
tomrobertsOF: I see you have a comment below, but could you add another quick one here saying why these are commented out.

src/provider/model/ZIndexer.ts
```diff
@@ -97,7 +97,7 @@ export class ZIndexer {
             } else if (identity.uuid === SERVICE_IDENTITY.uuid && !identity.name.startsWith('TABSET-')) {
                 // Exclude service-owned windows
                 return false;
-            } else if (exclude && exclude.uuid === identity.uuid && exclude.name === identity.name) {
+            } else if (exclusions.find(exclusion => exclusion.uuid === identity.uuid && exclusion.name === identity.name)) {
```
tomrobertsOF: If this is only being used as a condition, could we use `Array.some(...)` instead since it returns a boolean.

src/provider/model/ZIndexer.ts
bryangaleOF: 👍

src/provider/model/ZIndexer.ts
bryangaleOF: 👍

src/provider/model/ZIndexer.ts
bryangaleOF: Ah, good spot, fixed.

src/provider/model/ZIndexer.ts
tomrobertsOF: Thanks!

PR #314, opened 2019-02-27
[SERVICE-359] Improve input guards on client methods

tomrobertsOF:
- Added some additional checks to the `parseIdentity` checks, specifically around the types of the arguments provided.
- Made a new `parseIdentityRule` method which does a similar thing but for identities allowing Regex.

Note: I've also included a fix for an error I encountered during this story where the undock methods weren't properly returning provider-side errors to the client.

### DIFF
--- a/src/client/internal.ts
+++ b/src/client/internal.ts
@@ -9,7 +9,7 @@
  * This file is excluded from the public-facing TypeScript documentation.
  */
 import {Identity} from 'hadouken-js-adapter';
-import {WindowIdentity} from './main';
+import {WindowIdentity, IdentityRule, RegEx} from './main';
 import {ApplicationUIConfig, TabProperties} from './tabbing';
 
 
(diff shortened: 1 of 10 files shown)


src/client/internal.ts
```diff
@@ -50,10 +50,47 @@ export function getId(): WindowIdentity {
  */
 export function parseIdentity(identity: WindowIdentity|Identity): WindowIdentity {
     if (!identity || !identity.uuid) {
-        throw new Error('Invalid Identity provided.  A valid Identity contains both a uuid and name');
+        throw new Error('Invalid Identity provided: A valid Identity contains both a uuid and name');
```
bryangaleOF: Just want to check that this error message matches what we're checking for in this block. What are we concerned identity.uuid might be? null/undefined/empty string? Would checking for `!identity.uuid` be better rolled into uuidCheck?

src/client/internal.ts
```diff
@@ -50,10 +50,47 @@ export function getId(): WindowIdentity {
...
+
+    return {uuid: identity.uuid, name: identity.name || identity.uuid};
+}
 
+/**
+ * Like parseIdentity above, but also allows properties to be regex objects.
+ */
+export function parseIdentityRule(identity: IdentityRule): IdentityRule {
+    if (!identity || !identity.uuid) {
+        throw new Error('Invalid Identity provided: A valid Identity contains both a uuid and name');
```
bryangaleOF: Ditto above

src/client/internal.ts
tomrobertsOF: This is basically just checking for the existence of the uuid (and also hasn't been changed in this PR other than chaining a "." to a ":" and removing an extra space. .

bryangaleOF: Yep, understand all that. Just wondered if it'd be neater to drop the `identity.uuid` check from this line 52, modify the error in line 53, and change line 56 to something `const uuidCheck = !!identity.uuid && typeof identity.uuid === 'string';`. Then in all cases the error message will reference the specific field(s) that caused the error, and we're doing our uuid checking all in one line, just like we're doing our name checking all in one line.

tomrobertsOF: That's a fair point. Honestly, I was just trying to leave as much as possible as it was since a fair amount of time was spent going through the whole client last week. Will see what I can get done today.

src/client/main.ts
```diff
@@ -101,7 +101,7 @@ export async function deregister(identity: IdentityRule = getId() as IdentityRul
     if (!identity.uuid || !identity.name) {
         throw new Error('Invalid window identity provided');
```
MichaelMCoates: Does this becomes redundant with the check below?

src/client/main.ts
```diff
@@ -133,5 +133,5 @@ export async function register(identity: IdentityRule = getId() as IdentityRule)
...
```
MichaelMCoates: Same as above.

src/client/main.ts
tomrobertsOF: Yea, probably.

tomrobertsOF: Fixed.

src/client/internal.ts
```diff
@@ -49,13 +50,48 @@ export function getId(): WindowIdentity {
  * Assumed that the supplied object has uuid & name.
  */
 export function parseIdentity(identity: WindowIdentity|Identity): WindowIdentity {
-    if (!identity || !identity.uuid) {
-        throw new Error('Invalid Identity provided.  A valid Identity contains both a uuid and name');
+    if (identity === null || typeof identity !== 'object') {
+        throw new Error(ErrorMsgs.IDENTITY_REQUIRED);
+    }
+    const uuidCheck = typeof identity.uuid === 'string';
```
bryangaleOF: The previous check would have failed if identity.uuid was an empty string, but now it won't. Are we worried about that? Content if the answer is no, but want to check it's been considered.

test/client/TabApi.unittest.ts
```diff
@@ -47,15 +47,22 @@ describe('Tabbing API Actions', () => {
 
     describe('When calling removeTab', () => {
```
bryangaleOF: Worth still testing the uuid = null/name = null cases?

src/client/internal.ts
tomrobertsOF: Hmmm. Thing is there's problems with just checking truthiness here. An empty string would just return an error from the provider along the lines of "window not found", which i think is fine.

test/client/TabApi.unittest.ts
tomrobertsOF: It still does for some of the other api calls. Since they're all really testing the same code I figured I would do a bit of a mix-and-match.

tomrobertsOF:
Waiting for Jenkins before merging

PR #318, opened 2019-02-28
[SERVICE-328, SERVICE-377] Updated API docs, removed need for v1 types within client

pjbroadbent:
Updated workspaces API docs, and some minor tweaks/clean-up elsewhere.

Some code changes to clean-up `Workspace`/`WorkspaceWindow`, and to get the service to stick to those definitions (removed some unused fields that were being added to these objects during generate calls).

### DIFF
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -14,10 +14,10 @@
 import {EventEmitter} from 'events';
 import {ChannelClient} from 'hadouken-js-adapter/out/types/src/api/interappbus/channel/client';
 
-import {Events as SnapAndDockEvents} from '../client/snapanddock';
-import {Events as TabbingEvents} from '../client/tabbing';
-import {Events as TabstripEvents} from '../client/tabstrip';
-import {Events as WorkspacesEvents} from '../client/workspaces';
+import {SnapAndDockEvent} from '../client/snapanddock';
+import {TabbingEvent} from '../client/tabbing';
+import {TabstripEvent} from '../client/tabstrip';
+import {WorkspacesEvent} from '../client/workspaces';
 
 import {APITopic, SERVICE_CHANNEL} from './internal';
 
@@ -31,7 +31,7 @@ declare const PACKAGE_VERSION: string;
 /**
  * Defines all events that are fired by the service
  */
-export type Events = TabstripEvents|TabbingEvents|WorkspacesEvents|SnapAndDockEvents;
+export type LayoutsEvent = TabstripEvent|TabbingEvent|WorkspacesEvent|SnapAndDockEvent;
 
 /**
  * The event emitter to emit events received from the service.  All addEventListeners will tap into this.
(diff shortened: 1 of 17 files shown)


src/client/tabbing.ts
```diff
@@ -366,10 +370,12 @@ export async function setActiveTab(identity: Identity = getId()): Promise<void>
  * // Closes another windows context tab.
  * tabbing.closeTab({uuid: 'App1', name: 'App1'});
  * ```
+ * 
+ * Note that the `identity` doesn't need to be tabbed for this call to take effect - so long as the window is known with the service.
```
bryangaleOF: Worth saying something like "this will work for untabbed windows but we recommend you use {whatever you should use} for that"?

bryangaleOF: Or "For untabbed windows this will behave as {foo}"

src/client/workspaces.ts
```diff
@@ -116,19 +117,24 @@ export interface WorkspaceApp {
...
-export interface WorkspaceWindow extends Bounds, WindowIdentity {
+export interface WorkspaceWindow extends WindowIdentity {
+    /**
+     * The full URL of the window, corresponds to the `url` property of `Window.getInfo()`
+     */
+    url: string;
+
     /**
-     * If the window is currently visible, corresponds to `Window.isShowing()`.
+     * If the window is currently visible, corresponds to `Window.isShowing()`
```
bryangaleOF: What's our rule on fullstops?

src/provider/workspaces/utils.ts
```diff
@@ -245,24 +255,27 @@ export function parseVersionString(versionString: string): SemVer {
...
             const applicationState = tabWindow.applicationState;
             const tabGroup = tabWindow.tabGroup;
             if (tabGroup) {
                 const tabStripHeight = tabGroup.config.height;
+                const bounds = layoutWindow.bounds;
 
-                layoutWindow.top = layoutWindow.top - tabStripHeight;
-                layoutWindow.height = layoutWindow.height + tabStripHeight;
+                bounds.top = bounds.top - tabStripHeight;
+                bounds.height = bounds.height + tabStripHeight;
```
bryangaleOF: Use `+=` and `-=` consistent with the below change?

pjbroadbent:
Have added SERVICE-377 to this PR too. @tomrobertsOF noticed that we only have that issue due to a single reference to `fin.ApplicationOptions`. Have replaced that type with `object` (which is what the v2 types use for `ApplicationOptions`) and instead linked to best-available application options docs within the tsdoc comment for that field.

src/client/workspaces.ts
pjbroadbent: Not sure there is one...

Removed this, as the convention _seems_ to be "full stops only if the comment is multiple sentences".

src/client/main.ts
```diff
@@ -36,7 +36,14 @@ export interface WindowIdentity {
...
 export interface IdentityRule {
+    /**
+     * Application identifier
+     */
     uuid: string|RegEx;
+
+    /**
+     * Window identifier
+     */
     name: string|RegEx;
```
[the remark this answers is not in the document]
pjbroadbent: Not particularly obvious that `name` acts as a window ID, but is a part of standard OpenFin API conventions. We're best sticking to those conventions than renaming to something different.

src/client/connection.ts
```diff
@@ -23,6 +28,11 @@ import {APITopic, SERVICE_CHANNEL} from './internal';
  */
 declare const PACKAGE_VERSION: string;
 
+/**
+ * Defines all events that are fired by the service
+ */
+export type Events = TabstripEvents|TabbingEvents|WorkspacesEvents|SnapAndDockEvents;
```
tomrobertsOF: I vaguely remember a previous PR where we agreed that type names should generally be singular

PR #320, opened 2019-02-28
[N/A] Fix Loading... title for Placeholder Windows in tabs

MichaelMCoates:


src/provider/model/DesktopWindow.ts
```diff
@@ -172,6 +172,11 @@ export class DesktopWindow implements DesktopEntity {
                     }
                 }
 
+                // Special handling for workspace placeholder windows
+                if (options && options.name && options.name.startsWith('Placeholder-')) {
+                    windowTitle = 'Loading...';
+                }
+
```
bryangaleOF: Should we also check that the uuid is 'layouts-service'?

src/provider/model/DesktopWindow.ts
```diff
@@ -172,6 +172,11 @@ export class DesktopWindow implements DesktopEntity {
                     }
                 }
 
+                // Special handling for workspace placeholder windows
+                if (options && options.name && options.name.startsWith('Placeholder-')) {
```
pjbroadbent: This is clearly a special-case for just _our_ placeholder windows, can we check `identity.uuid` and `identity.name`, rather than `options.name`?

MichaelMCoates: gotcha

PR #326, opened 2019-03-01
[N/A] Modify ZIndexer to reduce test failure rate

bryangaleOF:
This partially addresses SERVICE-380, but does leave some so-far unexplained random test failures remaining.

Many of the random test failures were caused by `tabWindowsTogether` not resulting in windows being tabbed together, which was ultimately caused by `ZIndexer` occasionally not properly tracking an OF window.

I also saw some cases locally where the topmost window in the stack would be the preview window, which seemed to be another cause of test failures, but I've been unable to confirm this occured on Jenkins.

This PR
- Adds logging to `tabWindowsTogether` so we can track when it does not work as expected. That is complicated by the fact that occasionally we do not expect `tabWindowsTogether` to actually tab windows together
- Changes `ZIndexer` to add windows to the stack ASAP upon creation, rather than waiting for that window's first `shown` or `focused` event.
- Changes to `ZIndexer` to filter service windows from both `getTopMost` and `getWindowAt`

### DIFF
--- a/src/provider/model/ZIndexer.ts
+++ b/src/provider/model/ZIndexer.ts
@@ -79,7 +79,7 @@ export class ZIndexer {
             if (item.active) {
                 const index: number = ids.indexOf(item.id);
                 if (index >= 0) {
-                    if (!this.isPreviewWindow(item)) {
+                    if (!this.isServiceWindow(item)) {
                         return items[index];
                     } else {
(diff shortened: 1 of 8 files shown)


bryangaleOF:
Intentionally not adding reviewers yet

MichaelMCoates:
Passed CLI

PR #327, opened 2019-03-05
[SERVICE-384] Prevent reattachment of a DesktopWindow that no longer exists

MichaelMCoates:
Original bug was calling `bringToFront` on windows that do not exist, which threw the error `bringToFront is not a function`. The bug was caused by reattachment of a DesktopWindow that doesn't exist (when closing out a TabSet that is snapped to other windows).

I added a check to the reattachment code to only attach the DesktopWindow if it is `ready`. Is that the most dependable call to know if a window exists? Would we be better served adding the `ready` check in `setSnapGroup` instead?

Same goes for the `bringToFront` check. Should we use the `ready` check there instead?

### DIFF
--- a/src/provider/model/DesktopTabGroup.ts
+++ b/src/provider/model/DesktopTabGroup.ts
@@ -442,8 +442,10 @@ export class DesktopTabGroup implements DesktopEntity {
                 // the other snapped windows.
                 // TODO (SERVICE-311): Investigate how to properly harden against these issues
                 await new Promise(res => setTimeout(res, 10));
-                console.log('Re-attaching remaining tab: ' + remainingTab.id + ' => ' + joinedSnappable.id);
-                await remainingTab.setSnapGroup(joinedSnappable.snapGroup);
+                if (remainingTab.isReady) {
+                    console.log('Re-attaching remaining tab: ' + remainingTab.id + ' => ' + joinedSnappable.id);
+                    await remainingTab.setSnapGroup(joinedSnappable.snapGroup);
+                }
             }
         }
     }
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -764,10 +764,10 @@ export class DesktopWindow implements DesktopEntity {
                         // Bring other windows in group to front
                         await windows.map(groupWindow => {
                             // Checks to make sure the window actually exists before bringing it to front. Fix for SERVICE-384
-                            if (groupWindow._window && groupWindow._window.bringToFront) {
+                            if (groupWindow.isReady) {
                                 groupWindow._window.bringToFront();
                             } else {
-                                console.warn("groupWindow._window.bringToFront called with a nonexistent window: ", groupWindow);
+                                console.warn("groupWindow is not ready. You may have a nonexistent window in your snapGroup: ", groupWindow);
                             }
                         });
                     }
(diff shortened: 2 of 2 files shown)


src/provider/model/DesktopWindow.ts
```diff
@@ -762,7 +762,14 @@ export class DesktopWindow implements DesktopEntity {
                         const windows: DesktopWindow[] = this._snapGroup.windows as DesktopWindow[];
 
                         // Bring other windows in group to front
-                        await windows.map(groupWindow => groupWindow._window.bringToFront());
+                        await windows.map(groupWindow => {
+                            // Checks to make sure the window actually exists before bringing it to front. Fix for SERVICE-384
+                            if (groupWindow._window && groupWindow._window.bringToFront) {
```
bryangaleOF: I feel like I may have misunderstood something, but given the groupWindow._window is a class instance, it'd be very odd for `groupWindow._window.bringToFront` to not exist, right?

bryangaleOF: Also, `_window` is assigned in the `DesktopWindow` constructor, and nowhere else. I'm not seeing where or how it can become `null` or `undefined`.

MichaelMCoates: So I dug a bit deeper, and you're correct, `_window` can't become null or undefined. However, `_window.bringToFront` ceases to exist once the OpenFin _Window has `close()` called on it.

MichaelMCoates: Which happens in the `DesktopTabGroup.closeAll()` -> `DesktopTabGroup.removeAllTabs(true)`

MichaelMCoates: Should I remove the first `_window` check?

bryangaleOF: My feeling is yes

bryangaleOF: Also, should we be worried that a non-existent window is in `this._snapGroup.windows` at all? Does this leave us with a snapgroup with a lingering non-existent window in it's `windows` list? Is there a more fundamental fix that makes sure the non-existent window is removed from the list before we get to this line?

src/provider/model/DesktopTabGroup.ts
```diff
@@ -442,8 +442,10 @@ export class DesktopTabGroup implements DesktopEntity {
                 // the other snapped windows.
                 // TODO (SERVICE-311): Investigate how to properly harden against these issues
                 await new Promise(res => setTimeout(res, 10));
-                console.log('Re-attaching remaining tab: ' + remainingTab.id + ' => ' + joinedSnappable.id);
-                await remainingTab.setSnapGroup(joinedSnappable.snapGroup);
+                if (remainingTab.isReady) {
```
bryangaleOF: Is it weird that a non-ready tab is in our `_tabs` list at all at this point? (Or has it already been removed and we're just hanging onto it from before the timeout?)

src/provider/model/DesktopWindow.ts
MichaelMCoates: Ooooh, right, that may get us where we want, I'll test it out. 

Yes, that's the fix I made with the re-attachment. The re-attachment was the cause of the lingering non-existent window, this is the symptom.

src/provider/model/DesktopTabGroup.ts
MichaelMCoates: Correct, it has been removed and we're just hanging onto it. So we need to make sure that it didn't close while this function was running.

src/provider/model/DesktopWindow.ts
MichaelMCoates: `isReady` check works great!

MichaelMCoates:
Correct, that's exactly it. I'll leave the latter in for safety/logging to see if there's another place in the code that causes this error in the future.

PR #330, opened 2019-03-08
[SERVICE-365] Allow tabstrip maximise/restore by double-clicking an empty portion of the tabstrip window

bryangaleOF:
**Edit: Not ready for review. Discovered issue where this approach breaks snapping for tabgroups**

Previously we used the `webkit-app-region: drag` CSS attribute to allow users to drag the tabstrip window. However, this swallowed all other DOM mouse events¹, preventing us from listening for the double click event.

To solve this, this PR removes the use of `webkit-app-region: drag`, and instead listens for mouse up and down events, and moves the window around using the OpenFin API. This also allows us to easily prevent dragging while the tabstrip is maximized, as a stopgap solution until we can address SERVICE-369.

One downside of this approach is that we lose the context menu with move, maximize, minimize, etc. options that came for free with `webkit-app-region: drag`. We also lose "shake window to clear desktop" on this window, and any other OS-level feature than came or will come for free with `webkit-app-region: drag`.

¹ https://github.com/electron/electron/issues/1354 (I've tried the solution suggested in the last comment, but it did not appear to work)

### DIFF
--- a/res/demo/tabstrips/custom1.css
+++ b/res/demo/tabstrips/custom1.css
@@ -55,7 +55,6 @@ body {
 	justify-content: space-between;
 	margin-left: -15px;
 	min-width: 72px;
-	-webkit-app-region: no-drag;
 }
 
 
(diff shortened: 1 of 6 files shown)


bryangaleOF:
Closing due to no sensible workaround for issues with this approach

PR #331, opened 2019-03-08
[SERVICE-369] The TabStrip is movable when in a maximized state.

bryangaleOF:
**Edit: Not ready for review. Discovered issue where this approach breaks snapping for tabgroups**

This builds on PR-328.

When DesktopTabWindow detects the tabstrip is being moved while maximized, it restores the window. This required some modification in main.ts, relative to PR-328 to support the window size changing during a drag operation.

I think the worst part of this PR, is that the provider makes a guess about when the window is being dragged from mouse interaction. An alternative would be for the tabstrip to explicitly make a restoreWhileDragging call (or add a new boolean to the existing restore API method), but I wanted to avoid API additions if possible. Would appreciate any thoughts if I should go with this approach instead.

### DIFF
--- a/res/demo/tabstrips/custom1.css
+++ b/res/demo/tabstrips/custom1.css
@@ -55,7 +55,6 @@ body {
 	justify-content: space-between;
 	margin-left: -15px;
 	min-width: 72px;
-	-webkit-app-region: no-drag;
 }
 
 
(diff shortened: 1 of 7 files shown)


bryangaleOF:
Closing due to dependence on closed https://github.com/HadoukenIO/layouts-service/pull/330

PR #334, opened 2019-03-12
[SERVICE-402] Add an undock button to the default tabstrip

bryangaleOF:
The only subtlety is the change in DesktopSnapGroup, which fixes an issue with events silently not being sent because the window wasn't yet `_ready`.


src/provider/model/DesktopSnapGroup.ts
```diff
@@ -437,4 +431,18 @@ export class DesktopSnapGroup {
...
         }
     }
+
+    private async sendEventIfReadyOrCheckLater(window: DesktopWindow, type: 'window-docked'|'window-undocked'): Promise<void> {
+        if (!window.isReady) {
+            await window.sync();
+        }
+
+        if (window.isReady) {
+            if (type === 'window-docked') {
```
pjbroadbent: Can you not avoid the if-else here? Just `window.sendEvent({type});`?

src/provider/model/DesktopSnapGroup.ts
```diff
@@ -437,4 +431,18 @@ export class DesktopSnapGroup {
             };
         }
     }
+
+    private async sendEventIfReadyOrCheckLater(window: DesktopWindow, type: 'window-docked'|'window-undocked'): Promise<void> {
```
pjbroadbent: IMO, this is window-level behaviour. Seems to me that this kind of robustness check should just be a part of `DesktopWindow.sendEvent`.

bryangaleOF: Since I missed a couple of places in `DesktopSnapGroup` that send events but aren't going through this, I'm convinced :p

src/provider/model/DesktopSnapGroup.ts
bryangaleOF: Push stuff down into DesktopWindow, so no longer an issue

src/provider/model/DesktopSnapGroup.ts
bryangaleOF: Done

PR #335, opened 2019-03-13
[SERVICE-403] Disable tabbing previews when tabbing feature is disabled

bryangaleOF:
This story leaves it open how we should handle the currently-dragged window having tabbing disabled. For this PR, I've taken the view that for the preview window to show, both the target and active windows should have tabbing enabled.

Much of change of logic gets lost where I've reshuffled the contents of TabService.ts. so suggest looking specifically at the first and third commits in this PR.


src/provider/tabbing/TabService.ts
```diff
@@ -339,18 +339,33 @@ export class TabService {
      * @param item2 Details about the second entity, either window about to be tabbed or the tabstrip a window is to be added to
      */
     public canTabTogether(item1: DesktopEntity, item2: DesktopEntity): boolean {
-        const configs: {enabled: boolean; config: Tabstrip}[] = [item1, item2].map(entity => {
+        return this.isEntityTabEnabled(item1) && this.isEntityTabEnabled(item1) && this.doEntitiesShareTabstripConfig(item1, item2);
```
MichaelMCoates: Is the second check here supposed to be `item2` instead of `item1`?

src/provider/tabbing/TabService.ts
```diff
@@ -469,7 +486,8 @@ export class TabService {
             const targetTabGroup = targetWindow.tabGroup;
 
             // Check if the target and active window have same tab config.
-            const valid = this.constraintsCompatible(activeWindow, targetTabGroup || targetWindow) && this.canTabTogether(targetWindow, activeWindow);
+            const valid =
+                this.constraintsCompatible(activeWindow, targetTabGroup || targetWindow) && this.doEntitiesShareTabstripConfig(targetWindow, activeWindow);
```
MichaelMCoates: Is there a reason why we check the `TabGroup` at all here? Is there a way for the `targetWindow` and the `targetTabGroup` to get out of sync somehow?

src/provider/tabbing/TabService.ts
bryangaleOF: Yes! Fixed!

src/provider/tabbing/TabService.ts
bryangaleOF: `targetTabGroup` might not exist yet. If we're dragging one untabbed window over another untabbed window, `targetTabGroup` will be `null`, so we go off `targetWindow`.

Does that resolve this?

PR #341, opened 2019-03-20
[SERVICE-400] Re-integrate extracted tooling back into Layouts project

ryanpetersonOF:
This integrates the `openfin-service-tooling` NPM package into the layouts project.  This will move building and serving out of the project and let the package handle it.  With this approach we can keep a consistent build/serve process across all of the services project.  This project is the first to receive it so some bumpy waters may be expected, at first.

There is a current exception for testing which for now will continue to use the local express, but does make use of the package middlewares.


bryangaleOF:
@ryanpetersonOF Kicked off Jenkins for this, but not expecting any miracles. Happy to squerge if this has passed locally for you

ryanpetersonOF:
@bryangaleOF It failed Jenkins yesterday with one unrelated error, and is passing locally for me.  Let's squerge it

PR #342, opened 2019-03-20
[SERVICE-380][SERVICE-210] Jestify Layouts Tests

bryangaleOF:
This PR converts existing tests to use Jest rather than Ava, which allows us to automatically retry failing tests (see `setup.js`)

This was largely a fairly mechanical transformation, though I want to draw attention to
- Changes in `connect.ts` - Unlike Ava, each testsuite in Jest seems to be well sandboxed, so there is no way to share this connection object between each testsuite. There's also no equivalent disconnect method to clean stuff up nicely, so each testsuite has to use a different name.

- There may be some further worthwhile work to better name the tests, so they display nicer within Jenkins. This would involve wrapping each testsuite in a `describe` block, which, would obscure the meaningful changes in this PR


test/demo/utils/workspacesUtils.ts
```diff
@@ -70,21 +65,21 @@ async function getTestApps(): Promise<Application[]> {
...
 }
 
-export async function createCloseAndRestoreLayout(t: TestContext|SaveRestoreTestContext): Promise<Workspace> {
+export async function createCloseAndRestoreLayout(context: AppContext|undefined = undefined): Promise<Workspace> {
     const workspace = await sendServiceMessage(WorkspaceAPI.GENERATE_LAYOUT, undefined) as Workspace;
 
-    assertIsLayoutObject(t, workspace);
-    if (isSaveRestoreContext(t)) {
+    assertIsLayoutObject(workspace);
+    if (context !== undefined && isSaveRestoreContext(context)) {
```
MichaelMCoates: I think this first `context !== undefined` check is covered in `isSaveRestoreContext`

test/demo/snapanddock/advancedSnapAndDock.test.ts
```diff
@@ -8,15 +8,15 @@ import {teardown} from '../../teardown';
...
 
 // Using testParameterized more for type safety than parameterization
 // since this just one very specific test
 testParameterized(
     (testOptions: CreateWindowData) => 'Cannot snap windows so they overlap - shape: U',
     [{windowCount: 5, frame: true}],
-    createWindowTest(async (t, testOptions: CreateWindowData) => {
-        const windows = t.context.windows;
+    createWindowTest(async (context, testOptions: CreateWindowData) => {
+        const windows = context.windows;
```
MichaelMCoates: Purely stylistic thing, some of these tests use `const windows = context.windows` and others use `{windows} = context`. Should we align on one?

test/demo/utils/workspacesUtils.ts
bryangaleOF: Simplified this to get rid of isSaveRestoreContext, as that should be sorted for us by the type system

test/demo/snapanddock/advancedSnapAndDock.test.ts
bryangaleOF: Yes, but will cause massive conflicts with PR-352. Will see if I can do this easily as part of that

test/demo/utils/workspacesUtils.ts
MichaelMCoates: 👍

test/demo/snapanddock/advancedSnapAndDock.test.ts
MichaelMCoates: 👍

test/provider/deregisterConfig.test.ts
```diff
@@ -30,52 +36,53 @@ async function addRuleToProvider(scope: Scope, config: ConfigurationObject): Pro
...
     windows.push(await createChildWindow({...DEFAULT_OPTIONS, name: 'testWindow2'}));
 
     await dragSideToSide(windows[0], 'left', windows[1], 'right');
-    await assertGrouped(t, ...windows);
+    await assertGrouped(...windows);
     await addRuleToProvider({level: 'window', uuid: 'testApp', name: 'testWindow1'}, {enabled: false});
 
-    t.false(await isWindowRegistered(windows[0].identity));
-    t.true(await isWindowRegistered(windows[1].identity));
+    assert.strictEqual(await isWindowRegistered(windows[0].identity), false);
```
MichaelMCoates: Everything else seems to pass, but this check seems to fail sometimes. Adding a delay before seems to resolve it.

package.json
```diff
@@ -29,23 +29,28 @@
...
-    "jest-cli": "^23.6.0",
-    "jest-junit": "^5.2.0",
+    "jest": "^24.5.0",
+    "jest-circus": "^24.5.0",
+    "jest-cli": "^24.5.0",
+    "jest-junit": "^6.3.0",
     "json-schema-defaults": "^0.4.0",
     "json-schema-to-typescript": "^6.1.0",
     "mkdirp": "^0.5.1",
+    "power-assert": "^1.6.1",
```
tomrobertsOF: Why do we need a new assertion package? Are the built-in jest assertions not good enough?

test/client/TabApi.unittest.ts
```diff
@@ -10,7 +10,7 @@ import {tabbing, WindowIdentity} from '../../src/client/main';
 stub();
 
 let channel: ChannelClient;
-let channelDispatch: jest.SpyInstance<typeof channel.dispatch>;
+let channelDispatch: jest.SpyInstance<Promise<any>, [string, any?]>;
```
tomrobertsOF: Is there not some better typing we could do here rather than just `Promise<any>`?

test/demo/utils/createWindowTest.ts
```diff
@@ -25,28 +24,29 @@ export interface WindowContext {
     windowInitializer: WindowInitializer;
 }
 
-export function createWindowTest<T extends CreateWindowData, C extends WindowContext = WindowContext>(
-    testFunc: TestMacro<T, C>, windowOptions?: fin.WindowOptions, customArrangements?: ArrangementsType, customWindowPositions?: WindowPosition[]):
-    TestMacro<T, C> {
+export function createWindowTest<T extends CreateWindowData>(
```
tomrobertsOF: Why did we remove the ability to extend the default context?

test/provider/store.unittest.ts
```diff
@@ -371,8 +371,8 @@ describe('Store', () => {
         const config = {bool: true};
 
         let watch: ScopeWatch<Config>;
-        let onAddProxy: jest.Mock<WatchCallback>;
-        let onRemoveProxy: jest.Mock<WatchCallback>;
+        let onAddProxy: jest.Mock<void, any>;
```
tomrobertsOF: Can these be something other than any?

package.json
bryangaleOF: Lots of layouts tests contain something like assertFoo(...., "Some message about what the assertion failure means");. Switching to Jest-style asserts would have meant throwing the additional message away

test/demo/utils/createWindowTest.ts
bryangaleOF: It wasn't being used, and with the change to Jest which doesn't involve passing around a TestContext, it fell away naturally.

A couple of alternatives would be:
- Have the caller pass in an empty object that we then attach window and window initializer to, rather than creating `{windows, windowInitializer}` as a new object. This felt clumsy as right now, it'd involve every test using createWindowTest passing in an empty object (or having it as a default param) and I'd need to add `|null` to the fields in `WindowContext` to get the object to initialize before the windows are attached.
- Have the caller pass in an additional object that goes through to the actual test. Having an extra param feels clumsy, and right now this'd just be an empty object for everything.

Neither of those felt nicer than leaving up to the caller, if they want additional context going into the tests, to set that up in `beforeEach` and keep it in the test file

test/demo/utils/workspacesUtils.ts
```diff
@@ -70,25 +51,25 @@ async function getTestApps(): Promise<Application[]> {
                            }));
 }
 
-export async function createCloseAndRestoreLayout(t: TestContext|SaveRestoreTestContext): Promise<Workspace> {
+export async function createCloseAndRestoreLayout(context: AppContext|undefined = undefined): Promise<Workspace> {
```
pjbroadbent: Declare this as `createCloseAndRestoreLayout(context?: AppContext)`?

test/demo/workspaces/basicSaveAndRestore.test.ts
```diff
@@ -26,22 +24,22 @@ numberOfApps.forEach(appNumber => {
...
+afterEach(teardown);
 
-testParameterized<CreateAppData, AppContext>(
+testParameterized<CreateAppData>(
     (testOptions: CreateAppData): string => `Basic SaveAndRestore - ${testOptions.apps[0].createType === 'manifest' ? 'Manifest' : 'Programmatic'} - ${
         testOptions.apps.length} App(s) - ${testOptions.apps[0].childWindows.length} Child(ren) Each`,
     basicTestOptionsArray,
-    createAppTest(async (t, applicationData: CreateAppData) => {
-        await createCloseAndRestoreLayout(t);
+    createAppTest(async (context, applicationData: CreateAppData) => {
```
pjbroadbent: Is there a type that can be added to these `context` arguments?

`async (context: ???, applicationData: CreateAppData) => {`

test/provider/deregisterConfig.test.ts
pjbroadbent: The `addRuleToProvider` util currently has no way to await all the side-effects of a config change.

Adding a small delay inside `addRuleToProvider` may solve some race conditions.

test/provider/store.unittest.ts
pjbroadbent: `jest.Mock<void, [Store<Config>, ScopedConfig<Config>]>`?..

test/client/TabApi.unittest.ts
bryangaleOF: It's less descriptive than it was, but it's correct given that dispatch takes `[string, any]` and returns `Promise(any)`

We could do `jest.SpyInstance<ReturnType<typeof channel.dispatch>, Parameters<typeof channel.dispatch>>`, which would be equivalent but more descriptive. Unfortunately, jest-typescript chokes on `Parameters<>` as it seems to be fairly new

test/provider/store.unittest.ts
bryangaleOF: Fixed, but wish there was some way to do a `SignalArgs` type, so we could do `jest.Mock<void, SignalArgs<typeof ScopeWatch.prototype.onAdd>>`. I gave it a try, but not even sure if it's possible. Shout if you know the typescript magic to do this.

test/demo/workspaces/basicSaveAndRestore.test.ts
bryangaleOF: It's being correctly inferred as an `AppContext`. Is it that we want to make it explicit, or is it that the function signature is mixing inferred and explicit typing? If it's the former, is there a rule of thumb for when we went to make these things explicit?

package.json
tomrobertsOF: Ok, we should probably look at this more closely down the line. Problem is more that we are have tests written in the style/format of one test framework but running them in a different framework with different standard patterns for these kinds of things.

Long-term we should probably actually "jestify" the whole test-suite, i.e. restructure it to be more inline with jest's structure/patterns.

test/demo/utils/createWindowTest.ts
tomrobertsOF: Can be revisited if/when we do our proper test restructuring

test/provider/store.unittest.ts
tomrobertsOF: Haven't tried it myself, but you should be able to reference property types like `ScopeWatch['onAdd']`, assuming that onAdd is definitely typed at compile-time

bryangaleOF: It's not so much referencing the property type, it's finding some way to go from `Signal2<A, B>` to `[A, B]`

PR #348, opened 2019-03-25
[N/A] ZIndexer fix

bryangaleOF:
This follows on from PR-342. Only `ZIndexer.ts` has changed compared to PR-342

ZIndexer used to rely on timestamp to order its entities. If two events arrive in the same millisecond, this could give unexpected results. This was sometimes visible with `tabToMaximizedWindow.test.ts`, which did the following:

```
await windows[2].setAsForeground();
await windows[1].setAsForeground();
await windows[0].setAsForeground();
```

which when relying on timestamp, may not give the expected results, giving a random, performance dependent, test failure. This PR changes `ZIndexer` to use a counter rather than timestamp


MichaelMCoates:
Was the issue before that the `(timestamp < entry.timestamp)` would fail? It would be equal instead?

bryangaleOF:
@MichaelMCoates No, it was that when two entities had the same timestamp, we had no way of storing which window was on top of the other. It made the result of `sortStack` unpredictable

pjbroadbent:
Changes look good, but will leave this till after #342 is merged.

PR #350, opened 2019-03-25
[NA] Update README.md

ryanpetersonOF:
Removes references to server.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ To setup the project whilst avoiding the above dependencies, can instead be inst
 ### Startup
 Once dependencies are installed, start the "built-in" sample application with `npm start`. This uses `webpack-dev-middleware` to both build and host the application; a custom server script will start the OpenFin application once the server is up and running.
 
-The startup script has optional arguments which can be used to tweak the behavior of the build and the test server. See the constants at the top of [this file](https://github.com/HadoukenIO/service-tooling/blob/develop/src/server/server.ts) for details on the available parameters and their effects.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,9 +102,9 @@ After checkout, install project dependencies using `npm install`. The integratio
 To setup the project whilst avoiding the above dependencies, can instead be installed with `npm install --ignore-scripts`.
 
 ### Startup
-Once dependencies are installed, start the "built-in" sample application with `npm start`. This uses `webpack-dev-middleware` to both build and host the application; a custom `server.js` script will start the OpenFin application once the server is up and running.
+Once dependencies are installed, start the "built-in" sample application with `npm start`. This uses `webpack-dev-middleware` to both build and host the application; a custom server script will start the OpenFin application once the server is up and running.
 
-The startup script has optional arguments which can be used to tweak the behavior of the build and the test server. See the constants at the top of `server.js` for details on the available parameters and their effects.
+The startup script has optional arguments which can be used to tweak the behavior of the build and the test server. See the constants at the top of [this file](https://github.com/HadoukenIO/service-tooling/blob/develop/src/server/server.ts) for details on the available parameters and their effects.
```
pjbroadbent: There are no constants in `server.ts`. The arguments are documented in https://github.com/HadoukenIO/service-tooling/blob/develop/src/types.ts.

Would it be better to say "See `npm start -- -h`"?

ryanpetersonOF: I think it was the `CLIArguments` but it is much better to just use the `--h`

PR #351, opened 2019-03-25
[SERVICE-401] Fix "unhandled promise rejection" errors when including client from non-OpenFin contexts

MichaelMCoates:
Phil's implementation. After a lot of messing around, couldn't figure out a way to optimize it further. 

Essentially fetches the channel client from a function instead of an awaited promise.

### DIFF
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -41,24 +41,33 @@ export const eventEmitter = new EventEmitter();
 /**
  * Promise to the channel object that allows us to connect to the client
  */
-export const channelPromise: Promise<ChannelClient> = typeof fin === 'undefined' ?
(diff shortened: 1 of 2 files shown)


bryangaleOF:
I just want to make sure I understand the change properly

Old behaviour:
Regardless of the context, we'd try to create the channel promise immediately, leading to errors in non-OpenFin contexts

New behaviour:
We try to create the channel promise immediately, __only if we're in an OpenFin context__. Otherwise, we only create it on-demand (so we'll still see errors in non-OpenFin contexts, but only if `getServicePromise()` is actually called.

Is that correct?

MichaelMCoates:
Correct!

PR #355, opened 2019-03-28
[NA] Fix Client Connection

ryanpetersonOF:
Fixes the client/connection.ts error prevalent in the demo

```
channelPromise is undefined
```

### DIFF
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -41,7 +41,7 @@ export const eventEmitter = new EventEmitter();
 /**
  * Promise to the channel object that allows us to connect to the client
  */
-let channelPromise: Promise<ChannelClient>|null;
+let channelPromise: Promise<ChannelClient>|null = null;
 channelPromise = (typeof fin !== 'undefined') ? getServicePromise() : null;
 
 export function getServicePromise(): Promise<ChannelClient> {
--- a/src/client/connection.ts
+++ b/src/client/connection.ts
@@ -41,8 +41,8 @@ export const eventEmitter = new EventEmitter();
 /**
  * Promise to the channel object that allows us to connect to the client
  */
-
-let channelPromise: Promise<ChannelClient>|null = (typeof fin !== 'undefined') ? getServicePromise() : null;
+let channelPromise: Promise<ChannelClient>|null;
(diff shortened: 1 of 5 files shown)


bryangaleOF:
I'm confused how this makes a difference to anything!

src/client/connection.ts
```diff
@@ -41,8 +41,8 @@ export const eventEmitter = new EventEmitter();
 /**
  * Promise to the channel object that allows us to connect to the client
  */
-
-let channelPromise: Promise<ChannelClient>|null = (typeof fin !== 'undefined') ? getServicePromise() : null;
+let channelPromise: Promise<ChannelClient>|null;
+channelPromise = (typeof fin !== 'undefined') ? getServicePromise() : null;
```
bryangaleOF: Feel like this would be clearer as 

```
let channelPromise: Promise<ChannelClient>|null = null;
if (typeof fin !== 'undefined') {
    getServicePromise();
}
```

ryanpetersonOF: You still need to do

```
let channelPromise: Promise<ChannelClient>|null = null;
if (typeof fin !== 'undefined') {
    channelPromise = getServicePromise();
}
```

ryanpetersonOF: The fin check is also already in `getServicePromise`

bryangaleOF: getServicePromise() will do the assignment though?

bryangaleOF: And in `getServicePromise()`, failing the `typeof fin !== 'undefined'` test causes a rejected Promise, so we need to do it separately outside `getServicePromise()` since we want to avoid the rejected Promise in cases where the user hasn't explicitly called `getServicePromise()`.

src/client/connection.ts
```diff
@@ -41,8 +41,11 @@ export const eventEmitter = new EventEmitter();
 /**
  * Promise to the channel object that allows us to connect to the client
  */
+let channelPromise: Promise<ChannelClient>|null = null;
```
pjbroadbent: Unrelated, but something else that needs fixing. The unit tests still attempt to import `channelPromise` (which is now no longer exported), can we change those to import `getServicePromise()`?

MichaelMCoates: Good catch! Changed.

PR #364, opened 2019-04-04
RUN-5108 add external windows as they are being created

whyn07m3:


src/provider/model/DesktopModel.ts
```diff
@@ -77,7 +77,20 @@ export class DesktopModel {
             this.addIfEnabled({uuid: evt.uuid, name: evt.name});
         });
 
-        // Register any windows created before the service started
+        // Listen for external windows shown event and register only new ones.
+        // Using "...shown" event here instead of "...created" because most external
```
tomrobertsOF: Not sure how relevant it may be to native windows, but when tabbing we hide the non-active tabs, and show them when made active, so you could end up re-registering a load of windows if tabbing without some kind of guard in-place

whyn07m3: I'm using a lookup on line 86 and 88 to only register new native windows.
Is it sufficient or is there another window map I should check before registering?

tomrobertsOF: Ah, I hadn't looked very closely. I remember we had some extra handling in the normal registration for if we got a deregister during some of the async parts, but this will probably be fine for a first pass.

PR #365, opened 2019-04-04
[N/A] Add code examples for workspaces

MichaelMCoates:


src/client/workspaces.ts
```diff
@@ -334,6 +334,17 @@ export function removeEventListener<K extends WorkspacesEvent>(eventType: K['typ
  *
  * The callback will be invoked on each call to {@link generate}, and the return value (if anything is returned)
  * will be saved as the workspace's `customData` property for this app within the generated {@link Workspace}.
+ * 
+ * ``` ts
+ * // import the client module
+ * import * as Layouts from "openfin-layouts"
```
pjbroadbent: Can we align the imports with the other code samples?

Remove the comment, and import using `import {foo} from 'openfin-layouts'`.

src/client/workspaces.ts
```diff
@@ -410,6 +462,17 @@ export async function generate(): Promise<Workspace> {
  * service will then group all windows that were formerly snapped together.
  *
  * Finally, the layouts service will send a 'workspace-restored' event to all windows, and complete restoration.
+ * 
+ * ``` ts
+ * //import the client module
+ * import * as Layouts from "openfin-layouts"
+ * 
+ * Layouts.workspaces.restore(workspaceObject).then(result => {
+ *    //promise resolves with result once the layout has been restored
```
pjbroadbent: Super minor, but match the comment style to that used in the service/demo code, and the rest of the code samples?

`// Promise resolves...`

src/client/workspaces.ts
```diff
@@ -357,6 +368,34 @@ export async function setGenerateHandler(customDataDecorator: () => CustomData):
...
+ *     const ofApp = await fin.Application.getCurrent();
+ *     const openWindows = await ofApp.getChildWindows();
+ *     //iterate through the child windows of the workspaceApp data
+ *     const opened = workspaceApp.childWindows.map(async (win, index) => {
+ *         //check for existence of the window
+ *         if (!openWindows.some(w => w.identity.name === win.name)) {
+ *             //create the OpenFin window with the same name
+ *             const ofWin = await openChild(win.name, win.info.url);
+ *             //position the window based on the data in the workspaceApp
+ *             await positionWindow(win);
```
bryangaleOF: Could we comment that `positionWindow` is something the user has to provide themselves? And I know it wouldn't strictly need to, but I'd find it clearer if positionWindow took in both the existing openfin window and the workspace so make it clear it's making one match the other. (So suggest do openWindows.find instead of openWindows.some, and have one call to positionWindow outside the if-else block)

(And observation not strictly related to the PR, looking at the demo apps, *wow* there's a lot of detail any user has to get precisely right to properly restore child windows)

src/client/workspaces.ts
```diff
@@ -334,6 +334,17 @@ export function removeEventListener<K extends WorkspacesEvent>(eventType: K['typ
...
  * The callback will be invoked on each call to {@link generate}, and the return value (if anything is returned)
  * will be saved as the workspace's `customData` property for this app within the generated {@link Workspace}.
+ * 
+ * ``` ts
+ * // import the client module
+ * import * as Layouts from "openfin-layouts"
+ * 
+ * Layouts.workspaces.setGenerateHandler(() => {
+ *     // return custom data
+ *     return {"foo":"bar"};
```
bryangaleOF: Could this be an example of real world usage? `return {currentStockSymbol: this._currentStockSymbol}`?

(https://dev.to/warrend/no-more-foo-bar-and-baz-56j4 - last few sentences are relevant)

src/client/workspaces.ts
MichaelMCoates: Will do!

src/client/workspaces.ts
MichaelMCoates: 👍

src/client/workspaces.ts
MichaelMCoates: Will do! Pushed up, is that what you were thinking? 

Separately, we can also consider providing the `positionWindow` function as part of the client code. Could make the child window restoration a bit easier.

src/client/workspaces.ts
MichaelMCoates: Great idea!

MichaelMCoates:
> Looks good. Concur with Phil that we should make comments match our own style, and also suggest that the code should satisfy our own linter

Doing a linting pass now.

src/client/workspaces.ts
```diff
@@ -410,6 +460,16 @@ export async function generate(): Promise<Workspace> {
  * service will then group all windows that were formerly snapped together.
  *
  * Finally, the layouts service will send a 'workspace-restored' event to all windows, and complete restoration.
+ * 
+ * ``` ts
+ * import {workspaces} from 'openfin-layouts';
+ * 
+ * workspaces.restore(workspaceObject).then(result => {
+ *    // Promise resolves with result once the layout has been restored
+ *    handleResult(result)
```
bryangaleOF: Semi-colon

MichaelMCoates: 👍

src/client/workspaces.ts
```diff
@@ -375,6 +412,19 @@ export async function setRestoreHandler(listener: (workspaceApp: WorkspaceApp) =
  * If an application wishes to restore its child windows and store custom data, it must properly integrate with the
  * layouts service by both registering {@link setGenerateHandler|generate} and {@link setRestoreHandler|restore} callbacks, and
  * calling the {@link ready} function. If this is not done properly, workspace restoration may be disrupted.
+ * 
+ * ``` ts
+ * import {workspaces} from 'openfin-layouts';
+ * import {Workspace} from 'openfin-layouts/dist/client/workspaces';
+ *
+ * async function saveCurrentWorkspace {
```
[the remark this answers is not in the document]
MichaelMCoates: Ahhh good catch, adding the (). 

I think it's best to keep it in, just to help drive down the point that the layouts-service doesn't store anything, so storage is in the developer's hands.

PR #367, opened 2019-04-04
[SERVICE-423] Renamed global variable created by non-module client

pjbroadbent:
Renamed from `OpenFinLayouts` global variable to `layouts`.

This is to align with the same change made in FDC3, and to make this version of the client less clunky to use.

See https://github.com/HadoukenIO/fdc3-service/pull/30.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,9 @@ The client library is also available as a resource which can be included via `<s
 ```
    <script src="https://cdn.openfin/services/openfin/layouts/<VERSION>/openfin-layouts.js"></script>
 ```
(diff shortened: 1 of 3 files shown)


pjbroadbent:
> Assume arguments around populating the global namespace with things _not_ prefixed with OpenFin have been fully had?
> 
> We should also change the reference to OpenFinLayouts in res/demo/libScriptIncluded.html

Updated `libScriptIncluded.html`.

Yeah, was a discussion item in the last sync call. IMO since you're opting-in by including the script within your page, then it's a reasonable enough way to expose the functionality.

PR #370, opened 2019-04-08
[SERVICE-415] Fix 'restore' button on maximized window not working as expected when tabgroup was created from maximized window

bryangaleOF:
When creating a maximized tab group, the restore bounds of the original window are preserved, so the restore button on the tab group works as expected


src/provider/model/DesktopTabGroup.ts
```diff
@@ -235,10 +235,16 @@ export class DesktopTabGroup implements DesktopEntity {
                 await this.setSnapGroup(new DesktopSnapGroup());
             }
 
-            const {center, halfSize} = this._activeTab && this._activeTab.currentState || this._tabs[0].currentState;
 
-            this._beforeMaximizeBounds = {center: {...center}, halfSize: {...halfSize}};
+            if (beforeMaximizeBounds) {
```
pjbroadbent: Rather than the optional argument, can we keep everything self-contained in this function?

If this `if` were to be changed to `if (this._activeTab.currentState.state === 'maximized')`, would that have the same effect?

bryangaleOF: That wouldn't work, as by the time we get to calling maximize, then tab has already been normal-ed and resized. I've pushed an alternative approach, thought I'm not convinced it's preferable. More self-contained, but less explicit

src/provider/model/DesktopTabGroup.ts
```diff
@@ -321,6 +322,17 @@ export class DesktopTabGroup implements DesktopEntity {
         const firstTab: DesktopWindow = tabs.shift()!;
         const activeTab: DesktopWindow = (activeTabId && this._model.getWindow(activeTabId)) || firstTab;
 
+        // If we're forming this tabgroup from a maximized tab, we'll want to maximize this tabgroup, and inherit the tab window's restore bounds
```
MichaelMCoates: From what I can see, this only sets `beforeMaximizeBounds` if the first tab is maximized. What is the expected behavior if one of the other tabs is maximized instead?

bryangaleOF: Yep, in that case we would ignore the fact that the first tab is maximized. Generally we take the dimensions of the first tab to set the dimensions of the newly created tab group, so this is consistent behaviour

bryangaleOF:
Tests failed, but kicking off again as they took 2 hours, which makes me suspect something else was going on

PR #372, opened 2019-04-09
[N/A] Use a single test connection for Integration Tests

bryangaleOF:
Before:
Each testfile set up it's own connection to OpenFin. Since we had no way of closing this connection, we'd give each connection a random name. Connections would keep accumulating until all tests completed

After:
A single test connection is created using `integrationTestEnv.js` which is shared between all tests. This brings it in line with FDC3 tests


test/demo/utils/fin.ts
```diff
@@ -0,0 +1,4 @@
+import {Fin} from 'hadouken-js-adapter';
+
+declare const global: NodeJS.Global&{__FIN__: Fin};
+export const fin = global.__FIN__;
```
tomrobertsOF: Have you linted this? I think it complains if you don't have a newline at the end of the file.

test/demo/utils/jestGlobal.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Import this file anywhere you need the global puppeteer or openfin connections.
+ *
+ * We use a custom jest environment to avoid needing to connect to puppeteer/openfin in every test.
+ * The connection objects are mounted in the global scope for running tests, and as such we they
+ * need to be included in the global type definitons.
+ */
+declare module NodeJS {
+    interface Global {
+        __FIN__: import('hadouken-js-adapter').Fin;
```
tomrobertsOF: If the only place we reference this global is in fin.ts I'd rather not use the ambient declaration. We should be fine with just the explicit declaration in fin.ts and importing the fin object from there as needed.

test/demo/utils/serviceUtils.ts
```diff
@@ -46,12 +45,11 @@ export async function sendServiceMessage<T, R>(message: APITopic, payload: T): P
 }
 
 async function getChannelClient() {
-    const fin: Fin = await getConnection();
     return fin.InterApplicationBus.Channel.connect(SERVICE_CHANNEL);
 }
 
-export const layoutsClientPromise = getConnection().then(fin => {
+export const layoutsClientPromise = new Promise<typeof import('../../../src/client/main')>(res => {
```
tomrobertsOF: Since we no longer need to await the fin promise here this could be simplified to not need the wrapping promise. Something like: 
```typescript
Object.assign(global, {fin, PACKAGE_VERSION: "TEST-CLIENT"})
export const layoutsClientPromise = import('../../../src/main/client');
```

test/provider/defaultConfig.inttest.ts
```diff
@@ -22,7 +22,7 @@ const context: TestContext = {
 };
 
 beforeAll(async () => {
-    (global as NodeJS.Global & {fin: Fin}).fin = await getConnection();
+    (global as NodeJS.Global & {fin: Fin}).fin = fin;
```
tomrobertsOF: I know we use this pattern quite a lot in tests, but may be better to switch to 
```
Object.assign(global, {fin, ...})
```
in situations where we only need the global thing to exists in another file.

test/demo/utils/fin.ts
bryangaleOF: Yep, linted, and I can spot a bunch of testfiles that don't end with a newline. Maybe Ryan's branch is stricter?

tomrobertsOF: Ok, no worries as long as it passes a check.

PR #375, opened 2019-04-11
[SERVICE-398] Adding release channels to build script

pjbroadbent:
Added the `svc-tool channels` step to the build process for any stable (i.e. `master`) builds. This will create several variants of the provider's `app.json` manifest, each targeting a runtime release channel. e.g: `app.runtime-beta.json` - which is the same as the regular `app.json`, only running on the `beta` runtime.

PR also includes some clean-up of the Jenkinsfile script:
* We now use `agent none` at the top of the file, to prevent the job from occupying an additional executor unnecessarily, this could result in deadlock when there were many jobs running at once.
* There's no longer duplication between the staging and stable builds, and the script is in a format that should allow it to be directly copied between projects.
  * The service name is currently still hard-coded within the script, this will be changed once we get a plugin installed that allows the Jenkinsfile to read JSON files.


Jenkinsfile
```diff
@@ -1,16 +1,28 @@
...
+            }
+        }
+
+        stage('Test') {
             parallel {
                 stage('Unit Tests') {
                     agent { label 'linux-slave' }
                     steps {
-                        sh "npm i --ignore-scripts"
-                        sh "npm run generate"
```
bryangaleOF: Want to double check that we no longer need this line

Jenkinsfile
```diff
@@ -35,84 +47,94 @@ pipeline {
...
+    DIR_BUILD_ROOT = env.DSERVICE_S3_ROOT + SERVICE_NAME + '/'
+    DIR_BUILD_VERSION = DIR_BUILD_ROOT + BUILD_VERSION
+
+    DIR_DOCS_ROOT = env.DSERVICE_S3_ROOT_DOCS + SERVICE_NAME + '/'
+    DIR_DOCS_CHANNEL = DIR_DOCS_ROOT + CHANNEL
+    DIR_DOCS_VERSION = DIR_DOCS_ROOT + BUILD_VERSION
+}
+
+def buildProject() {
+    sh "npm run clean"
```
bryangaleOF: We don't need npm install as part of building anymore?

Jenkinsfile
pjbroadbent: Yeah - the `generate` (also) happens as a post-install step, so was previously running generate twice.

Jenkinsfile
pjbroadbent: The install happens above, when the unit tests are ran. But come to think of it, not sure we can rely on Jenkins _always_ re-using the same workspace. I'll dig in a bit, may re-add the install just to be sure.

bryangaleOF: Yeah, figured it was something like this. Intuitively felt fragile to rely on `npm i` happening in a previous step

pjbroadbent: Have re-added the `npm install`.

tomrobertsOF:
@pjbroadbent, not sure if you saw but you've got a test failure in CI. Looks like there's a messed up import somewhere:
```
[2019-04-29T12:15:33.499Z]     test/provider/store.unittest.ts:1:21 - error TS2307: Cannot find module '../../gen/provider/config/layouts-config'.
[2019-04-29T12:15:33.499Z] 
[2019-04-29T12:15:33.499Z]     1 import {Scope} from '../../gen/provider/config/layouts-config';
[2019-04-29T12:15:33.499Z]                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #382, opened 2019-04-18
[SERVICE-447]/[SERVICE-449] Unable to tab windows/Windows in snap group with resize constraints grow rapidly when dragged

bryangaleOF:
- Expose 'displayScaling' boolean from DesktopModel, so we can use DS mitigations selectively
- Manually compute the transform type when display scaling is active
- Disable resize constraints when a window is being dragged, so prevent window-growth bug

### DIFF
--- a/src/provider/model/DesktopSnapGroup.ts
+++ b/src/provider/model/DesktopSnapGroup.ts
@@ -78,14 +78,14 @@ export class DesktopSnapGroup {
 
     private _validateGroup: Debounced<() => void, DesktopSnapGroup, []>;
 
-    private _storedResizeConstraints: Map<string, Point<ResizeConstraint>>|null;
+    private _resizeConstraintsOverridden: boolean;
 
     constructor() {
         this._id = DesktopSnapGroup._nextId++;
         this._entities = [];
         this._windows = [];
         this.rootWindow = null;
-        this._storedResizeConstraints = null;
+        this._resizeConstraintsOverridden = false;
 
         const refreshFunc = this.calculateProperties.bind(this);
         this._localBounds = new CalculatedProperty<Rectangle>(refreshFunc);
@@ -189,31 +189,25 @@ export class DesktopSnapGroup {
      * scaling is enabled
      */
     public suspendResizeConstraints(): void {
-        if (this._windows.length > 2 && !this._storedResizeConstraints) {
+        if (this._windows.length > 2 && !this._resizeConstraintsOverridden) {
             const nullConstraint: ResizeConstraint = {resizableMin: true, resizableMax: true, minSize: 0, maxSize: Number.MAX_SAFE_INTEGER};
             const nullConstraints: Point<ResizeConstraint> = {x: nullConstraint, y: nullConstraint};
 
-            this._storedResizeConstraints = new Map<string, Point<ResizeConstraint>>();
-
             for (const window of this._windows) {
-                this._storedResizeConstraints.set(window.id, window.currentState.resizeConstraints);
-
-                window.applyProperties({resizeConstraints: nullConstraints});
+                window.applyOverride('resizeConstraints', nullConstraints);
             }
+
+            this._resizeConstraintsOverridden = true;
         }
     }
 
     public restoreResizeConstraints(): void {
-        if (this._storedResizeConstraints) {
+        if (this._resizeConstraintsOverridden) {
             for (const window of this._windows) {
-                if (this._storedResizeConstraints.has(window.id)) {
-                    const constraints = this._storedResizeConstraints.get(window.id)!;
-
-                    window.applyProperties({resizeConstraints: constraints});
-                }
+                window.resetOverride('resizeConstraints');
             }
 
-            this._storedResizeConstraints = null;
+            this._resizeConstraintsOverridden = false;
         }
     }
 
--- a/src/provider/model/DesktopSnapGroup.ts
+++ b/src/provider/model/DesktopSnapGroup.ts
@@ -78,14 +78,14 @@ export class DesktopSnapGroup {
 
     private _validateGroup: Debounced<() => void, DesktopSnapGroup, []>;
 
-    private _resizeConstraintsOverridden: boolean;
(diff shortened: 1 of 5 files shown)


src/provider/model/DesktopWindow.ts
```diff
@@ -1050,20 +1069,82 @@ export class DesktopWindow implements DesktopEntity {
...
+                } else {
+                    // Test that we're resizing from the min or max edge, otherwise we have a move rather than purely a resize
+                    const possibleCentersGivenResize = [
+                        (startBounds.center[axis] - startBounds.halfSize[axis]) + halfSize[axis],
+                        (startBounds.center[axis] + startBounds.halfSize[axis]) - halfSize[axis]
+                    ];
+
+                    if (!possibleCentersGivenResize.some(possibleCenter => Math.abs(possibleCenter - center[axis]) <= MINIMUM_MOVE_CHANGE)) {
+                        move = true;
+                    }
```
pjbroadbent: Seems the intention here is to check if the centre of the rect is where it "should be" if this were a "true" resize operation. Since we know which way the rectangle is being resized, could this be done without the array and `some`?

```
const expectedCenterIfResize = startBounds.center[axis] + ((halfSize[axis] - startBounds.halfSize[axis]) * 0.5);
move = Math.abs(center[axis] - expectedCenterIfResize[axis]) >= MINIMUM_MOVE_SIZE;
```

src/provider/model/DesktopWindow.ts
```diff
@@ -1029,13 +1039,22 @@ export class DesktopWindow implements DesktopEntity {
         const halfSize: Point = {x: bounds.width / 2, y: bounds.height / 2};
         const center: Point = {x: bounds.left + halfSize.x, y: bounds.top + halfSize.y};
 
+        let transformType: Mask<eTransformType>;
+
+        if (this._userInitiatedBoundsChange) {
+            transformType = this.updateTransformType(event, this._userInitiatedBoundsChange);
```
pjbroadbent: Can this be merged into the `if` block below, or is the ordering of `updateTransformType` and `updateState` significant?

src/provider/model/DesktopSnapGroup.ts
```diff
@@ -181,6 +184,39 @@ export class DesktopSnapGroup {
...
+            const nullConstraints: Point<ResizeConstraint> = {x: nullConstraint, y: nullConstraint};
+
+            this._resizeConstraintsOverridden = true;
+
+            for (const window of this._windows) {
+                // We refresh here, otherwise we may not know about constraint changes made by the app via the runtime API, which
+                // would prevent applyOverride properly unsetting them
+                window.refresh().then(() => {
+                    if (this._resizeConstraintsOverridden) {
+                        window.applyOverride('resizeConstraints', nullConstraints);
```
pjbroadbent: With this change, is it still possible to maximize application windows after they have been tabbed and un-tabbed?

Might sound like a weird question, but this may trigger a known issue with how the runtime handles window constraints that we ran into pre-1.0.

src/provider/model/DesktopSnapGroup.ts
```diff
@@ -181,6 +184,39 @@ export class DesktopSnapGroup {
...
+        }
+    }
+
+    public restoreResizeConstraints(): void {
+        if (this._resizeConstraintsOverridden) {
+            for (const window of this._windows) {
+                window.resetOverride('resizeConstraints');
+            }
+
+            this._resizeConstraintsOverridden = false;
```
pjbroadbent: Could we add a `window.hasOverride` method, rather than using a flag for this?

src/provider/model/DesktopTabGroup.ts
```diff
@@ -192,8 +192,20 @@ export class DesktopTabGroup implements DesktopEntity {
...
-        return this.activeTab.applyOffset(offset, adjustedHalfSize);
+
+        const activeTabHalfSize: Point|undefined = halfSize && {x: halfSize.x, y: halfSize.y - tabstripHalfHeight};
+        const tabstripHalfSize: Point|undefined = halfSize && {x: halfSize.x, y: tabstripHalfHeight};
+
+        /**
+         * We can't depend on the tabstrip having the appropriate resize constraints at this point, due to a race condition with our
+         * mitigations for display scaling issues in DesktopSnapGroup, so we manually move both the active tab, and the tabstrip itself,
+         * rather than taking the earlier approach of just moving the active tab and relying on grouping and resize constraints to make
+         * everything work as expected. We particularly run into this race condition when snapping a tabgroup to another window
```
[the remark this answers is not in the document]
bryangaleOF: "Assuming we're confident that this method also works in non-display scaling mode, when resize constraints are still active?" - I was, but you're making me less so. Will change

src/provider/model/DesktopSnapGroup.ts
bryangaleOF: I've renamed the flag to better convey that it isn't the same as a hypothetical `hasOverride`.

Consider the case where `restoreResizeConstraints` gets called between the call to `window.refesh()` above and the call to `window.applyOverride('resizeConstraints', nullConstraints);` to see why we need this

src/provider/model/DesktopWindow.ts
bryangaleOF: That _should_ be fine, but given the calls to SnapGroup coming from `updateTransformType` it feels less fragile to do those before we call `updateState`. Like, do we want suspend/restoreResizeConstraints call while the window has the new state or old state?

src/provider/model/DesktopWindow.ts
bryangaleOF: Revised the math

src/provider/model/DesktopSnapGroup.ts
bryangaleOF: Yep!

src/provider/model/DesktopWindow.ts
pjbroadbent: Yeah, ok - let's leave as-is.

PR #383, opened 2019-04-24
[SERVICE-451] Fix tab group splicing when maximized

tomrobertsOF:
Added some additional logic to handle focused to stop the "keep tab group at same z-index" stuff being blocked when the group is maximized.


src/provider/model/DesktopWindow.ts
```diff
@@ -1081,6 +1081,10 @@ export class DesktopWindow implements DesktopEntity {
             this._snapGroup.windows
                 .filter(snapGroupWindow => snapGroupWindow !== this && !snapGroupWindow.isMaximizedOrInMaximizedTab() && !snapGroupWindow.currentState.hidden)
                 .forEach(snapGroupWindow => snapGroupWindow.bringToFront());
+        } else if (this._tabGroup && this._tabGroup.state === 'maximized') {
```
bryangaleOF: Is the maximized check needed? Would also suggest a one line comment here, to communicate we now have these two distinct cases where we're bringing forward attached windows

tomrobertsOF: Not strictly required, but this is all so fragile I don't know if I want to touch it any more than needed

PR #387, opened 2019-04-26
[N/A] Track changes when using ActionOrigin.SERVICE_TEMPORARY

bryangaleOF:
Previously, when we used `applyOverride()` and `resetOverride()` on `DesktopWindow`, the store value to reset to would get lost if there was a call to `refresh()` between the apply and reset steps. I ran into this problem in the display scaling feature branch, as we call can `refresh()` around the same time we make a window translucent for snapping.

This was because when calling `updateState()` with `ActionOrigin.SERVICE_TEMPORARY`, we were not properly updating `this._modifiedState` as we were for `ActionOrigin.SERVICE`. This fix changes it so that we properly update `this._modifiedState` for both `ActionOrigin.SERVICE` and `ActionOrigin.SERVICE_TEMPORARY`.

### DIFF
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -13,6 +13,7 @@ import {Debounced} from '../snapanddock/utils/Debounced';
 import {isWin10} from '../snapanddock/utils/platform';
 import {Point} from '../snapanddock/utils/PointUtils';
 import {Rectangle} from '../snapanddock/utils/RectUtils';
+import {forEachProperty} from '../utils/iteration';
 
 import {DesktopEntity} from './DesktopEntity';
 import {DesktopModel} from './DesktopModel';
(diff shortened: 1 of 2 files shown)


src/provider/model/DesktopWindow.ts
```diff
@@ -617,17 +618,13 @@ export class DesktopWindow implements DesktopEntity {
...
-                            hasChanges = true;
-                        }
+                forEachProperty(state, (property) => {
+                    if (this.isModified(property, appState, state) &&
+                        (!this._modifiedState.hasOwnProperty(property) || this.isModified(property, this._modifiedState, state))) {
+                        modifications[property] = state[property];
+                        hasChanges = true;
                     }
-                }
+                });
```
bryangaleOF: The above change is unrelated to the fix, but since I introduced forEachProperty, it seemed like using here. Since `state` is an object we generate, and not one that comes from the runtime, it seems safe to skip the `hasOwnProperty` check

src/provider/model/DesktopWindow.ts
```diff
@@ -845,34 +844,34 @@ export class DesktopWindow implements DesktopEntity {
...
+        } else {
+            // Track which properties have been modified by the service
+            forEachProperty(delta, (property) => {
+                if (this.isModified(property, delta, this._applicationState)) {
+                    this._modifiedState[property] = delta[property];
+                } else {
+                    delete this._modifiedState[property];
+                }
+            });
         }
```
bryangaleOF: This block was previous run for all `origin`s, even `APPLICATION`, which I believe was a mistake, and that it only makes sense to run this for `SERVICE` and `SERVICE_TEMPORARY`

PR #389, opened 2019-04-29
[SERVICE-464] Fix "tearing in" not working on a second monitor of a different display scale

bryangaleOF:


bryangaleOF:
> I'm slightly concerned about changing what was previously a fully synchronous sequence of events into an unawaited promise. Have you given thought to whether there any other operations that could jump in the middle and cause issues?

I've added a check to prevent DragWindowHandler possibly dispatching `onDragOver` after `onDragDrop`. When this does occur, it means the service will be operating on a slightly out of date mouse position when `onDragDrop` is fired, but it occurred rarely, and if someone is moving the mouse when they release the drop, I don't think this is going to be noticeable.

Other than that, from the PoV of anything outside of DragWindowHandler, the event will just be slightly later than it would be otherwise, but it was an event that was triggered unpredictably from outside the service anyway.

PR #394, opened 2019-05-01
[SERVICE-435] Display scaling support for 150% and 200%

bryangaleOF:
This enables support for display scaling at 150% and 200%, and includes mitigations for issues that occur at this scales

Misreporting of moves vs. resizes, so we compute this ourselves
Grouped windows resizing when being moved when some of the group includes resize constraints, so we temporarily resize constraints during moves
Fuzziness when checking group validity, since windows drift slightly when being moved, and don't always go quite where we want them to when they are first snapped

### DIFF
--- a/src/provider/main.ts
+++ b/src/provider/main.ts
@@ -35,16 +35,20 @@ declare const window: Window&{
 
 fin.desktop.main(main);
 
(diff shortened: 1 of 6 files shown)


pjbroadbent:
Jenkins issue is blocking job being created for this PR. Merging with status check still pending.

PR #400, opened 2019-05-13
[SERVICE-472] Add special validation check for right-click-moves

tomrobertsOF:
Certain bounds-changed events will now trigger a group validate. This will ensure that if a window is moved by through the context menu move it will get properly ejected from the group.

This is intended as a temporary workaround on the layouts side until a full solution is found from the runtime/core side

### DIFF
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -1145,7 +1145,7 @@ export class DesktopWindow implements DesktopEntity {
         this.updateTransformType(event, this._userInitiatedBoundsChange);
 
         // Temporary extra validation to workaround runtime issues with windows right-click move
-        // TODO: Removed once runtime has better handling of this edge case ([JIRA Ref])
+        // TODO: Removed once runtime has better handling of this edge case (RUN-5074?)
         let disabled = false;
         const disabledListener = () => {
             disabled = true;
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -1155,7 +1155,7 @@ export class DesktopWindow implements DesktopEntity {
             if (!disabled && this._snapGroup.length > 1) {
                 if (this._tabGroup) {
                     this._tabGroup.validate();
-                } else if (this._snapGroup.length > 1) {
+                } else {
                     this._snapGroup.validate();
                 }
             }
--- a/src/provider/model/DesktopWindow.ts
+++ b/src/provider/model/DesktopWindow.ts
@@ -1143,6 +1143,24 @@ export class DesktopWindow implements DesktopEntity {
         this._userInitiatedBoundsChange = {startBounds, transformType: 0};
 
         this.updateTransformType(event, this._userInitiatedBoundsChange);
+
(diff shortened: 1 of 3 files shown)


src/provider/model/DesktopWindow.ts
```diff
@@ -1143,6 +1143,24 @@ export class DesktopWindow implements DesktopEntity {
...
+        let disabled = false;
+        const disabledListener = () => {
+            disabled = true;
+        };
+        this._window.once('disabled-frame-bounds-changing', disabledListener);
+        this._window.once('bounds-changed', () => {
+            if (!disabled && this._snapGroup.length > 1) {
+                if (this._tabGroup) {
+                    this._tabGroup.validate();
+                } else if (this._snapGroup.length > 1) {
```
bryangaleOF: Would get rid of this check as we've already checked this above

src/provider/model/DesktopWindow.ts
```diff
@@ -1143,6 +1143,24 @@ export class DesktopWindow implements DesktopEntity {
...
         this.updateTransformType(event, this._userInitiatedBoundsChange);
+
+        // Temporary extra validation to workaround runtime issues with windows right-click move
+        // TODO: Removed once runtime has better handling of this edge case (RUN-5074?)
+        let disabled = false;
+        const disabledListener = () => {
+            disabled = true;
+        };
+        this._window.once('disabled-frame-bounds-changing', disabledListener);
+        this._window.once('bounds-changed', () => {
```
bryangaleOF: I'm not sure this is a case of **either** we want or validate the DTG **or** validate the DSG (let's say we have a tabgroup in a bigger snapgroup). I think it's more like maybe we want to try to repair the relative tab/tabstrip positioning (if we're a DTG), and always want to validate the DSG. It'd get super race condition-y to call one directly after the other here, but maybe DTG.validateInternal should call DSG.validate()?

src/provider/model/DesktopWindow.ts
```diff
@@ -1143,6 +1143,24 @@ export class DesktopWindow implements DesktopEntity {
         this._userInitiatedBoundsChange = {startBounds, transformType: 0};
 
         this.updateTransformType(event, this._userInitiatedBoundsChange);
+
+        // Temporary extra validation to workaround runtime issues with windows right-click move
+        // TODO: Removed once runtime has better handling of this edge case (RUN-5074?)
```
bryangaleOF: RUN-5074 is about display scaling and the changeType, not sure of the link?

tomrobertsOF: @tim-dinsdale has said that this gets fixed as a side-effect of the scaling stuff

src/provider/model/DesktopWindow.ts
tomrobertsOF: Fair. Top one was added later and I hadn't noticed it was there twice. Will remove.

src/provider/model/DesktopWindow.ts
tomrobertsOF: I think it would be easier to have a chat about this offline

PR #402, opened 2019-05-20
[SERVICE-421] Remove duplicated app/window/manifest creation utils

tim-dinsdale:


tim-dinsdale:
This is not the whole thing - this is removing some of the differences between the two calls. Namely

1. create-manifest now no longer provides handy defaults to URI or UUID
2. the sizes are now standardised

Then, change the tests so they still work. The next PR should work on the changes that are required to add features to manifest, which are at least autoShow being configurable.

test/runner.js
```diff
@@ -131,23 +131,23 @@ async function serve() {
...
+            const {uuid, url, defaultTop, defaultLeft, defaultHeight, defaultWidth, config, autoShow} = req.query;
             const additionalServiceProperties = config ? {config: JSON.parse(config)} : {};
 
             // Create manifest (based upon demo app manifest)
             const {shortcut, services, ...baseManifest} = require('../res/demo/app.json');
             const manifest = {
                 ...baseManifest,
                 startup_app: {
-                    uuid: uuid || 'save-restore-test-app-' + Math.random().toString(36).substring(2),
-                    url: url || 'http://localhost:1337/test/saveRestoreTestingApp.html?deregistered=false',
```
bryangaleOF: I feel a little confused. Did the defaults never get used?

test/runner.js
```diff
@@ -131,23 +131,23 @@ async function serve() {
...
                     autoShow: autoShow || true,
                     saveWindowState: false,
-                    defaultTop: defaultTop ? JSON.parse(defaultTop): 100,
-                    defaultLeft: 100,
-                    defaultHeight: 225,
-                    defaultWidth: 225
+                    defaultTop: defaultTop ? JSON.parse(defaultTop): 605,
+                    defaultLeft: defaultLeft? JSON.parse(defaultLeft) : 860,
+                    defaultHeight: defaultHeight? JSON.parse(defaultHeight) : 605,
+                    defaultWidth: defaultWidth? JSON.parse(defaultWidth) : 860
```
bryangaleOF: Have these numbers come from anywhere? Do we want the left coordinate to be the same as the width? And would format as `foo ? x : y` rather than `foo? x : y` to match elsewhere in code

src/provider/tabbing/tabstrip/TabManager.ts
```diff
@@ -101,7 +101,7 @@ export class TabManager {
 
     /**
      * Un-sets the active tab
-     * @method unsetActiveTab Removes the active status from the current active Tab.
+     * @function unsetActiveTab Removes the active status from the current active Tab.
```
bryangaleOF: This is just to stop `npm run fix` giving us spurious changes, right? Worth just getting rid of the `@function` line? We don't use `@function` elsewhere and I don't think this line adds anything

tim-dinsdale: Correct. Sure, will do

test/runner.js
tim-dinsdale: The uuid never did. The URL did, in one place, and I am now calling it explicitly.

test/runner.js
tim-dinsdale: Yes, they are duped from here.

https://github.com/HadoukenIO/service-tooling/blob/develop/src/server/middleware.ts#L119

The idea is that by harmonising these params more drift can't happen if there are other calls to manifest rather than create-manifest

tim-dinsdale: And, ack on the formatting

src/provider/model/ZIndexer.ts
```diff
@@ -76,7 +76,7 @@ export class ZIndexer {
      * NOTE: Implementation will not return any item within the input that does not exist within the ZIndexer util.
      *
      * @param items Array of window or identity objects
-     * @return The top-most of the input items, when sorted by z-index
+     * @returns The top-most of the input items, when sorted by z-index
      */
```
tim-dinsdale: This is the fix to the returnsssss bug, btw, which I _almost_ feel bad for removing

PR #403, opened 2019-05-20
[SERVICE-444] Add Service Tooling Testing Integration

ryanpetersonOF:
Lets service-tooling handle the startup of Jest unit/int tests.


test/demo/tabbing/createTabFromMultipleWindows.inttest.ts
```diff
@@ -17,9 +17,9 @@ afterEach(async () => {
     await win1.close();
     await win2.close();
     fin.InterApplicationBus.removeAllListeners();
-});
 
-afterEach(teardown);
```
bryangaleOF: Does this make a difference or is this just a style change?

test/demo/workspaces/basicSaveAndRestore.inttest.ts
```diff
@@ -27,8 +27,10 @@ numberOfApps.forEach(appNumber => {
     });
 });
 
-afterEach(closeAllPreviews);
-afterEach(teardown);
+afterEach(async ()=>{
```
bryangaleOF: Would format as `async () => {` to match elsewhere. Ditto for the other `afterEach(async ()=>{`s

test/demo/tabbing/createTabFromMultipleWindows.inttest.ts
ryanpetersonOF: The two `afterEach` were causing issues so I put them all into one.

test/demo/workspaces/basicSaveAndRestore.inttest.ts
tomrobertsOF: I would have thought the linter would catch that?

ryanpetersonOF: Good catch

ryanpetersonOF:
Just waiting for tests to pass on this

PR #405, opened 2019-05-22
[SERVICE-471] Implement toggling/setting of opacity effects during previews

ryanpetersonOF:
Adds custom opacity configuration for applications during snapping/tabbing operations.  The opacity can differ depending on if the window is `active` or `target`.

Tests are a WIP, but as this is to go out Friday I figured I'd get a PR up for the core work.


src/demo/testbed/EventsUI.ts
```diff
@@ -112,6 +114,18 @@ export class EventsUI {
...
+            let parsedResult: string;
+            try {
+                parsedResult = JSON.stringify(item.result, null, 4);
+            } catch (e) {
+                parsedResult = `Non-serializable value: ${item.result}`;
+            }
+
+            const pre = document.createElement('pre');
+            pre.innerText = `    ${(parsedResult.replace(/\n/g, '\n    '))}`;
+            item.captionElement.innerHTML += `:${pre.outerHTML}`;
```
bryangaleOF: With the `+=`, this is very dependent on updateStatus being only called once with a result, right? Any way to make it less fragile?

test/demo/snapanddock/previewOpacitySnap.inttest.ts
```diff
@@ -0,0 +1,286 @@
...
+
+/**
+ * Creates a set of windows, one per argument.
+ *
+ * If argument is provided, it will be used as the windows snap preview configuration. If `undefined`, window will use
+ * default service config.
+ *
+ * @param configs Set of snap preview configs
+ */
+async function createWindowsWithConfig(...configs: (PreviewConfig|undefined)[]): Promise<_Window[]> {
```
tomrobertsOF: This is a great util. I can see it being really useful in other tests suites, so in my mind might make sense to move it into it's own file, but maybe something to tackle down the line

tomrobertsOF: Also (definitely more long term), it would be great to plug this into the app/window creation utils we use in the parameterised test stuff

src/demo/testbed/WindowsUI.ts
```diff
@@ -57,13 +61,22 @@ export class WindowsUI {
...
 
     constructor(elements: Elements) {
         this._elements = elements;
         this._defaultConfig = {};
         this._selectedRule = null;
         this._rules = [];
 
+        // Find current runtime version
+        this._currentRuntime = 'stable';
+        fin.Application.getCurrentSync().getInfo().then(info => {
```
bryangaleOF: Why not `fin.System.getVersion()`?

src/demo/testbed/WindowsUI.ts
```diff
@@ -57,13 +61,22 @@ export class WindowsUI {
...
+    private _currentRuntime: string;
 
     constructor(elements: Elements) {
         this._elements = elements;
         this._defaultConfig = {};
         this._selectedRule = null;
         this._rules = [];
 
+        // Find current runtime version
+        this._currentRuntime = 'stable';
```
bryangaleOF: Feels like it would be safer to give this an empty default, rather than a value that's probably wrong but looks reasonable?

test/demo/snapanddock/previewOpacitySnap.inttest.ts
```diff
@@ -0,0 +1,286 @@
...
+            expect(await windows[0].getOptions()).toHaveProperty('opacity', 0.1);
+            expect(await windows[1].getOptions()).toHaveProperty('opacity', 0.2);
+        });
+    });
+
+    describe('When a window declares custom targetOpacity of null', () => {
+        it('Keeps its preset opacity when target', async () => {
+            windows = await createWindowsWithConfig({targetOpacity: null}, undefined);
+
+            await windows[0].updateOptions({opacity: 0.5});
```
tomrobertsOF: It wouldn't make a huge difference for this test, but it may be worth triggering a refresh on the DesktopWindow after calling updateOptions to make sure the model is all up-to-date. I think there's a util in `modelUtils` which does this already.

test/provider/utils/tabWindowsHover.ts
```diff
@@ -0,0 +1,11 @@
+import {_Window} from 'hadouken-js-adapter/out/types/src/api/window/window';
+
```
tomrobertsOF: I thought we already had a "drag and hover" type util from the first set of tabbing preview tests, or did we just use `dragWindowAndHover`? 

Either way, the pattern we have on the snapping side was just adding a `dropWindow` optional param to the `dragSideToSide`. To be consistent could we do a similar thing in `tabWindowsTogether` instead of making it a separate util?

src/provider/model/ZIndexer.ts
```diff
@@ -76,7 +75,7 @@ export class ZIndexer {
      * NOTE: Implementation will not return any item within the input that does not exist within the ZIndexer util.
      *
      * @param items Array of window or identity objects
-     * @return The top-most of the input items, when sorted by z-index
+     * @returns The top-most of the input items, when sorted by z-index
```
bryangaleOF: Tim's made the same change in his PR

test/demo/tabbing/previewOpacityTab.inttest.ts
```diff
@@ -0,0 +1,189 @@
...
+        windows = await createWindowsWithConfig(config1, config2);
+
+        const targetIndex: number = (activeIndex + 1) % 2;
+        await tabWindowsHover(windows[targetIndex], windows[activeIndex]);
+    }
+
+
+    describe('And windows are using the default configuration', () => {
+        beforeEach(async () => {
+            await init();
```
bryangaleOF: Any value to this being inside the beforeEach rather than the test itself as with the rest of the tests?

test/demo/tabbing/previewOpacityTab.inttest.ts
```diff
@@ -0,0 +1,189 @@
...
+    robot.mouseToggle('up');
+
+    await Promise.all(windows.map(window => window.close()));
+    windows.length = 0;
+
+    await teardown();
+});
+
+describe('When two windows are moved within tabbing distance', () => {
+    async function init(config1?: PreviewConfig, config2?: PreviewConfig, activeIndex: number = 1) {
```
bryangaleOF: Can this have a more descriptive name?

test/demo/tabbing/previewOpacityTab.inttest.ts
```diff
@@ -0,0 +1,189 @@
+import robot from 'robotjs';
```
bryangaleOF: Since most of these tests follow the same structure, they could be a single parameterized tests, at the trade of off some readability. Thoughts on making these parameterized?

test/provider/utils/tabWindowsHover.ts
ryanpetersonOF: Yes that would make sense.  I will just update `tabWindowsTogether` util

test/demo/tabbing/previewOpacityTab.inttest.ts
ryanpetersonOF: Yes that is the plan.  I put these tests up just to get something up but they will be reduced shortly.  In fact, we can pretty much combine the tab and snap tests.

test/demo/tabbing/previewOpacityTab.inttest.ts
ryanpetersonOF: The init will be the same for each test in here, where other tests differ in some way.  This will get cleaned up with parameterization

res/demo/testbed/index.html
```diff
@@ -206,7 +212,9 @@ <h6 class="modal-application mt-3">Launch Type</h6>
                                 <label for="inputRuntime" class="col-sm-3 col-form-label">Runtime</label>
                                 <div class="col-sm-9">
                                     <select id="inputRuntime" class="form-control-sm">
-                                        <option selected>stable</option>
+                                        <option value="current" selected>Use Current (stable)</option>
```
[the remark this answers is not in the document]
pjbroadbent: If for whatever reason the update fails, the fall-back behaviour is to assume `stable`. The default here matches the default in `WindowsUI.ts`. The alternative to using `stable` as the fallback would be to either leave as undefined behaviour, or put in some error handling and messaging - which I felt was a bit overkill for this.

But true, taking a few "it's only the demo app" liberties here around the edge cases.

res/provider/config/layouts-config.schema.json
```diff
@@ -125,19 +125,36 @@
...
+            "required": [],
+            "additionalProperties": false
+        },
         "tabstrip": {
             "type": "object",
             "description": "Specifies a custom tabstrip to be used whenever two windows are tabbed together. Windows can only be tabbed if their configurations both allow it - they must use the same tabstrip URL, and both have the 'tab' feature enabled.",
             "properties": {
                 "url": {
                     "type": "string",
-                    "title": "Tabstrip URL",
```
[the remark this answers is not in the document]
pjbroadbent: Because if `"title"` is specified, the TypeScript generation would create a new `type` alias for this, rather than just documenting the property as being a member of type `string`.

Before (ish - from memory):
```
/**
 * Sets a custom tabstrip URL
 */
type TabstripURL = string;
// Plus same for TabstripHeight

/**
 * Specifies a custom tabstrip to be used whenever two windows are tabbed together. Windows can only be tabbed if their configurations both allow it - they must use the same tabstrip URL, and both have the 'tab' feature enabled.
 */
export interface Tabstrip {
  url: TabstripURL;
  height: TabstripHeight;
}
```

After:
```
/**
 * Specifies a custom tabstrip to be used whenever two windows are tabbed together. Windows can only be tabbed if their configurations both allow it - they must use the same tabstrip URL, and both have the 'tab' feature enabled.
 */
export interface Tabstrip {
  /**
   * Sets a custom tabstrip URL
   */
  url: string;
  /**
   * Sets the height of the custom tabstrip, in pixels
   */
  height: number;
}
```

The `TabstripURL`/`Height` types seemed superfluous - the latter structure was more the original intention. Given the title and description fields overlap, also makes the API docs a bit more readable. When both fields are specified, Typedoc will use `${title} - ${description)` to generate the property documentation.

src/demo/testbed/EventsUI.ts
pjbroadbent: With current usage, it will only ever be called once. `updateStatus` is private and only ever used as a promise callback.

Will harden if we add any more complex use-cases of this, but the event log is currently all pretty basic.

test/demo/snapanddock/previewOpacitySnap.inttest.ts
pjbroadbent: Exact same function is copy-pasted into the tab tests. It's still pretty use-case-specific, but would definitely make sense to move into a separate file just to avoid the duplication.

test/demo/snapanddock/previewOpacitySnap.inttest.ts
pjbroadbent: Some overlap with SERVICE-476 here, which should be coming fairly soon. May make some of those utils redundant.

tomrobertsOF: True, and I was thinking that, but until that stuff is in I think it makes sense to keep the model state accurate through a test run. Again, not really a big problem for this test but I'd rather be safe than spend a few hours troubleshooting something that turns out to be cause by this.

src/provider/model/ZIndexer.ts
pjbroadbent: Will resolve at merge time. Even if reverting this line, will likely still conflict.

res/demo/testbed/index.html
```diff
@@ -304,35 +312,138 @@ <h6 class="modal-application mt-3">Launch Type</h6>
...
+                                            <option>0.6</option>
+                                            <option>0.7</option>
+                                            <option>0.8</option>
+                                            <option>0.9</option>
+                                            <option value="1">1.0</option>
+                                        </select>
+                                    </div>
+                                </div>
+                                <div class="row">
+                                    <label for="configSnap" class="col-sm-3 col-form-label pr-0">
```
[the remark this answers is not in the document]
pjbroadbent: `activeOpacityTab`, actually. Have updated.

src/demo/testbed/WindowsUI.ts
pjbroadbent: See https://github.com/HadoukenIO/layouts-service/pull/405/files#r286911999

IMO, stable was a reasonable fallback, but can change if there's something more appropriate.

res/demo/testbed/index.html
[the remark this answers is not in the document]
pjbroadbent: Changed the "placeholder" text to avoid use of the word "current" before we know what the current runtime actually is.

test/demo/snapanddock/previewOpacitySnap.inttest.ts
ryanpetersonOF: Made this a util

test/demo/tabbing/previewOpacityTab.inttest.ts
ryanpetersonOF: Describing what the init is doing in function name style would be a bit of a doozy, so I have opted for putting a simple comment for each.

PR #410, opened 2019-06-03
[SERVICE-465] Preview window customisation from config

TheoOttahOF:


res/provider/config/layouts-config.schema.json
```diff
@@ -139,6 +139,37 @@
                     "type": ["number", "null"],
                     "description": "Controls the opacity of each window when another window is about to be snapped/tabbed to it. If set to null, the service will not modify the opacity of the window.",
                     "default": 0.8
+                },
+                "overlayValid": {
+                    "$ref": "#/definitions/overlay-preview",
+                    "description": "Configures the preview rectangle that is shown to indicate the action that is about to happen"
```
bryangaleOF: I'd definitely try to be as consistent as possible with terminology in user-facing documentation, so maybe something like:

"Configures the appearance of the preview overlay that is shown when a window is about to be snapped/tabbed to another window"

"Configures the appearance of the preview overlay that is shown when a window cannot be snapped/tabbed to another window"

"Options for customizing the preview overlay shown for snap and tab actions"

Maybe something else, but would try not to mix all of "overlay-preview", "preview rectangle", and "colored preview rectangle"

src/provider/Preview.ts
```diff
@@ -1,42 +1,80 @@
...
+    /**
+     * Border CSS
+     */
+    border?: string;
+}
 
 /**
  * Visual indicator of the current snap target.
  *
  * Will create colored rectangles based on the given group. Rectangle color will be set according to snap validity.
```
bryangaleOF: Would update to reflect we can customize the general appearance

src/provider/Preview.ts
```diff
@@ -1,42 +1,80 @@
...
+            config = query.tab;
+        }
+
+        if (target.valid) {
+            previewWindow = this._successPreviewWindow;
+            overlay = config.overlayValid;
+        } else {
+            previewWindow = this._failurePreviewWindow;
+            overlay = config.overlayInvalid;
+        }
```
bryangaleOF: Definitely a matter of personal taste, but suggest at least consider change this to use the ternary operator. In favour of going that route, this would become 3 lines and the 'let's all changed to `consts`

src/demo/testbed/View.ts
```diff
@@ -80,6 +80,19 @@ export interface Elements {
     targetOpacitySnap: HTMLSelectElement;
     activeOpacityTab: HTMLSelectElement;
     targetOpacityTab: HTMLSelectElement;
+    snap_overlayInvalid_background: HTMLInputElement;
```
bryangaleOF: Would add a one-line comment about why we're using `_` in the variable names. It's definitely something I'd be tempted to do "right click -> rename symbol" on :p

src/demo/testbed/WindowsUI.ts
```diff
@@ -17,13 +17,25 @@ interface ConfigData {
...
     tabstripHeight: number;
-    targetOpacitySnap: number|null;
-    activeOpacitySnap: number|null;
-    activeOpacityTab: number|null;
-    targetOpacityTab: number|null;
+    targetOpacitySnap: number | null;
+    activeOpacitySnap: number | null;
+    activeOpacityTab: number | null;
+    targetOpacityTab: number | null;
+    snap_overlayInvalid_background: string | null;
```
bryangaleOF: Similar comment here

src/provider/Preview.ts
```diff
@@ -1,42 +1,80 @@
...
      * The 'isValid' parameter determines the color of the rectangles, indicating if releasing the window will
      * successfully join a snap/tab group
+     * @param target The preview target.
      */
     public show(target: PreviewableTarget): void {
-        const previewWindow = target.valid ? this._successPreviewWindow : this._failurePreviewWindow;
+        let previewWindow: fin.OpenFinWindow;
+        let config: RequiredRecursive<PreviewConfig>;
+        let overlay: Required<Overlay>;
+        const query = this._config.query(target.activeWindow.scope).preview;
```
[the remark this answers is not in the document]
TheoOttahOF: I agree. 

Currently the dragged activeWindow's previews are used when tabbing into another group. Spoke to phil about this and he said we'll wait for feedback on how people want this to be resolved.

res/provider/config/layouts-config.schema.json
TheoOttahOF: Perhaps we should standardize naming of stuff.

configure/customize overlay -> custom styling
snap|tab -> action
preview window/rectange -> preview overlay

src/provider/Preview.ts
```diff
@@ -51,34 +142,36 @@ export class Preview {
...
             defaultHeight: defaultHalfSize.y * 2,
-            opacity: 0.8,
+            opacity: 0,
             minimizable: false,
             maximizable: false,
-            defaultTop: -1000,
-            defaultLeft: -1000,
+            defaultTop: -10000,
+            defaultLeft: -10000,
             showTaskbarIcon: false,
```
bryangaleOF: Why was -1000 not enough? Possibility of second monitor off to the left?

Though, if setting the opacity to 0 is enough when we're hiding the preview, why isn't that enough when we're initially creating the window?

src/provider/Preview.ts
```diff
@@ -1,48 +1,139 @@
...
      */
     public show(target: PreviewableTarget): void {
-        const previewWindow = target.valid ? this._successPreviewWindow : this._failurePreviewWindow;
+        const valid: Validity = target.valid ? Validity.VALID : Validity.INVALID;
+        const previewType = target.type.toLowerCase() as PreviewType;
+        const previewWindow: fin.OpenFinWindow = this._previewWindows[previewType][valid];
+        const scope: Scope = target.activeWindow.tabGroup ? target.activeWindow.tabGroup.activeTab.scope : target.activeWindow.scope;
+
+        const overlayKey = target.valid ? 'overlayValid' : 'overlayInvalid';
+        const query = this._config.query(scope).preview;
```
bryangaleOF: Suggest consider doing this in `applyStyles`, and in `_previewWindows` we store the desired opacity for each window. Then we do all our config querying at one time and place

src/provider/Preview.ts
```diff
@@ -1,48 +1,139 @@
...
+    private _config: ConfigStore;
+    private _lastScope!: Scope;
 
-    constructor() {
+    constructor(config: ConfigStore) {
         this._activeWindowPreview = null;
+        this._config = config;
+        const query = this._config.query(defaultScope).preview;
+
+        const previewTypes = Object.keys(query)
```
bryangaleOF: I'd be keen to have a second opinion on this, but I feel like this section is a lot for the sake of populating a 2x2 array. I can see the check and the extensibility this gives us, but I'm also always wary of stuff that'd break under a 'rename' refactoring, or break 'right click -> find all references'. As far as VSC can tell, Preview.tab isn't referenced anywhere

src/provider/Preview.ts
```diff
@@ -1,48 +1,139 @@
+import deepEqual from 'fast-deep-equal';
+
+import {PreviewConfig, Preview as PreviewProps, Scope} from '../../gen/provider/config/layouts-config';
```
bryangaleOF: PreviewConfig import isn't being used

src/provider/Preview.ts
```diff
@@ -1,48 +1,139 @@
...
+    }
+
+    private onTransform(activeGroup: DesktopSnapGroup, type: Mask<eTransformType>): void {
+        const activeWindow = activeGroup.windows[0];
+        const scope: Scope = activeWindow.tabGroup ? activeWindow.tabGroup.activeTab.scope : activeWindow.scope;
+        this.preload(scope);
+    }
+
+    private preload(scope: Scope): void {
+        if (deepEqual(this._lastScope, scope)) {
```
bryangaleOF: I _think_ this could just be `===`. Like, each window creates a unique scope object in it's constructor anyway

src/provider/Preview.ts
```diff
@@ -1,48 +1,139 @@
...
     }
 
     /**
      * Shows a rectangle that matches the snap/tab group target of a dragged window.
      *
      * The 'isValid' parameter determines the color of the rectangles, indicating if releasing the window will
      * successfully join a snap/tab group
+     * @param target The preview target.
      */
     public show(target: PreviewableTarget): void {
```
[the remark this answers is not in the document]
bryangaleOF: Also, I wonder if we want to update the window styling in here? Like, I can believe that 99% of the time, what gets through to here as the active target is what previously triggered `onTransform`, and so everything works correctly. But it's not something where you can just look at the code and reason that it'll always work as expected. There's a little bit of faith needed to believe that everything will work correctly

src/provider/Preview.ts
```diff
@@ -1,48 +1,139 @@
...
+            };
+        }, {}) as PreviewWindows;
 
-        this._successPreviewWindow = this.createWindow('successPreview', SUCCESS_PREVIEW_BACKGROUND_CSS);
-        this._failurePreviewWindow = this.createWindow('failurePreview', FAILURE_PREVIEW_BACKGROUND_CSS);
+        DesktopSnapGroup.onCreated.add(this.onCreated, this);
+    }
+
+    private onCreated(group: DesktopSnapGroup): void {
+        group.onTransform.add(this.onTransform, this);
```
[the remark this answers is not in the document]
TheoOttahOF: Good catch, added a onDestroy.remove

src/provider/Preview.ts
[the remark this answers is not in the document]
TheoOttahOF: I'll add a preload call inside show just incase of that 1%.

PR #414, opened 2019-06-26
[SERVICE-465] Preview window customisation options

TheoOttahOF:
- Custom CSS overlays for preview windows using `background` & `border` attributes.
- Layout config schema includes customisation options. 
- Demo app creation includes customisation options.
- Preview window testing utils.
- Preview window styling tests.
- Updated existing tests to handle the changes to preview windows.
- Prevent preview windows flickering on start up.

PR #415, opened 2019-06-28
[NA] Prevent window flicker on service startup

TheoOttahOF:
Stops `DragWindowManager` & `DesktopTabstripFactory` windows flickering on service startup.
Also fixes tabstrip pooling. Before tabstrips pooling was async and overwriting the pool array, leaving windows that were not being managed.

#414 Removes the flicker from preview windows.

### DIFF
--- a/src/provider/model/DesktopTabstripFactory.ts
+++ b/src/provider/model/DesktopTabstripFactory.ts
@@ -19,8 +19,6 @@ const DEFAULT_UI_URL = (() => {
     return providerLocation.replace('provider.html', 'tabbing/tabstrip/tabstrip.html');
 })();
 
-let VIRTUAL_SCREEN: DipRect;
-
 /**
  * Handles creation and pooling of Tab Group Windows
  */
@@ -129,12 +127,9 @@ export class DesktopTabstripFactory {
      * @param window The window to hide.
      */
     private async hideOffScreen(window: _Window){
-        if (!VIRTUAL_SCREEN){
-            const {virtualScreen} = await fin.System.getMonitorInfo();
-            VIRTUAL_SCREEN = virtualScreen;
-        }
+        const {virtualScreen} = await fin.System.getMonitorInfo();
         const {width, height} = await window.getBounds();
-        await window.showAt(VIRTUAL_SCREEN.left - width, VIRTUAL_SCREEN.top - height);
+        await window.showAt(virtualScreen.left - width, virtualScreen.top - height);
         await window.hide();
     }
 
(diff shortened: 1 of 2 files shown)


src/provider/model/DesktopTabstripFactory.ts
```diff
@@ -102,20 +106,34 @@ export class DesktopTabstripFactory {
...
 
+    /**
+     * Hide a window offscreen so it doesn't flicker on startup.
+     * @param window The window to hide.
+     */
+    private async hideOffScreen(window: _Window){
+        const {virtualScreen} = await fin.System.getMonitorInfo();
+        const {width, height} = await window.getBounds();
+        await window.showAt(virtualScreen.left - width, virtualScreen.top - height);
+        await window.hide();
```
bryangaleOF: Want to make sure I understand why we're doing things in this order. Discuss when you have time?

src/provider/tabbing/DragWindowManager.ts
```diff
@@ -116,6 +117,7 @@ export class DragWindowManager {
                     smallWindow: true
                 },
                 () => {
+                    this._window.showAt(virtualScreen.left - 100, virtualScreen.top - 100);
```
bryangaleOF: Why at -100 when for similar code in DesktopTabstripFactory.ts we'e going by the precise window width?

src/provider/model/DesktopTabstripFactory.ts
```diff
@@ -23,6 +23,8 @@ const DEFAULT_UI_URL = (() => {
  */
 export class DesktopTabstripFactory {
     public static readonly DEFAULT_CONFIG: Tabstrip = {url: DEFAULT_UI_URL, height: 60};
+    private static readonly POOL_MAX_SIZE: number = 3;
```
bryangaleOF: Suggest either change:

        this.createAndPool(DesktopTabstripFactory.DEFAULT_CONFIG);
        this.createAndPool(DesktopTabstripFactory.DEFAULT_CONFIG);
        this.createAndPool(DesktopTabstripFactory.DEFAULT_CONFIG);

So it's:

    for (let i = 0; i < POOL_MAX_SIZE; i++) {
            this.createAndPool(DesktopTabstripFactory.DEFAULT_CONFIG);
    }

Or have `createAndPool` always fill up a pool to it's max size

src/provider/tabbing/DragWindowManager.ts
bryangaleOF: Discussed - Theo will change to defaultWidth: 100

src/provider/model/DesktopTabstripFactory.ts
TheoOttahOF: Added a loop

src/provider/tabbing/DragWindowManager.ts
TheoOttahOF: This has been added with a comment on why

PR #416, opened 2019-07-11
[SERVICE-512] Extract config into separate module

ryanpetersonOF:
Takes all of the config loader logic and moves it into its own NPM package: `openfin-service-config`.  Two tests have also been moved into the package: `store.unittest` & `loadConfig.inttest`


test/teardown.ts
```diff
@@ -1,8 +1,9 @@
 import {Window} from 'hadouken-js-adapter';
 import {WindowInfo, WindowDetail} from 'hadouken-js-adapter/out/types/src/api/system/window';
+import {Scopes, ConfigWithRules, StoredConfig} from 'openfin-service-config';
```
TheoOttahOF: `ConfigWithRules` is unused

ryanpetersonOF: Thanks.  I have removed it now.

webpack.config.js
```diff
@@ -1,22 +1,27 @@
...
 
+/**
+ * Import config generator plugins from openfin-service-config
+ */
+const {SchemaToDefaultsPlugin} = require('openfin-service-config/plugins/SchemaToDefaultsPlugin');
+const {SchemaToTypeScriptPlugin} = require('openfin-service-config/plugins/SchemaToTypeScriptPlugin');
 /**
  * Import the webpack tools from openfin-service-tooling
  */
 const webpackTools = require('openfin-service-tooling').webpackTools;
```
pjbroadbent: These are just imports, can we drop the comments?

And use the same import style on both - either with the destructuring, or without

ryanpetersonOF: Dropped the comments and added destructure

PR #417, opened 2019-07-12
[SERVICE-535] Create restore helper function

tomrobertsOF:
Add four new functions to `client/workspaces`:
 - `defaultRestoreHandler`: A simple, all-in-one function that can be passed to `setRetoreHandler`. Will loop through the given workspace and create/position all windows as needed. 
 - `createOrPosition`: A helper function mirroring some of the code from `defaultRestoreHandler` which is intended to be used modularly when creating custom restore handlers.
 - `createChild`: Creates a child window with the position and window options specified in the workspace. Called by `defaultRestoreHandler` and `createOrPosition`, and could in principle be used to create more complex custom handlers.
- `positionWindow`: Similar to `createChild` but intended to be used for windows which are running at the time of restore. Will position the window and update its state to match the workspace. Also used by `defaultRestoreHandler` and `createOrPosition`

The new functions have doc comments but they are not very good (in my opinion). Would greatly appreciate editorial comments to improve them.

The code for these functions was largely pulled from the restore handlers we were using in the demo windows, with some general tidying applied. I've also gone through the demo/test window src files and updated them to use the default handler where it seemed appropriate.

Final note: in recation to a comment from Phil I went through and replaced some lingering instances of `await wrap` with `wrapSync`. Not strictly part of this story, but they were mostly in the code I was touching anyway.


src/client/workspaces.ts
```diff
@@ -496,3 +496,110 @@ export async function restore(payload: Workspace): Promise<Workspace> {
...
+            await ofWin.restore();
+        } else if (state === 'minimized') {
+            await ofWin.minimize();
+        } else if (state === 'maximized') {
+            await ofWin.maximize();
+        }
+
+        await ofWin.updateOptions({frame});
+    } catch (e) {
+        console.error('position window error: ' + e);
```
pjbroadbent: Need to think about how to handle errors in these handlers. This one suppresses errors (with a console log), but `createChild` will throw/reject.

Failure cases should also be called-out in the docs, IMO. Other than that, docs look good.

Let's discuss as a follow-up

src/client/workspaces.ts
```diff
@@ -496,3 +496,110 @@ export async function restore(payload: Workspace): Promise<Workspace> {
...
+    }
+}
+
+/**
+ * Given a workspaceWindow object for a non-open window, this will create the window, position it,
+ * and update its visibility, state and frame to fit with the restoring workspace.
+ *
+ * Called by the defaultRestoreHandler for any non-running windows in the workspace
+ */
+export async function createChild(workspaceWindow: WorkspaceWindow): Promise<void> {
```
pjbroadbent: Worth considering our terminology here, should this be `restoreChild`? Or something that outlines it as a restore handler util?

To that point, worth putting these in a separate file, or somehow separating these from the rest of the `workspaces` APIs? Would be worth generating the docs, and see how things look with these all mixed-in together.

src/client/restoreHelpers.ts
```diff
@@ -0,0 +1,115 @@
...
+        throw new Error('One or more errors encountered when restoring children: ' + errors.map(e => e.message).join(', '));
+    }
+}
+
+/**
+ * An additional helper function that can be used to create cutom restore handlers.
+ *
+ * Given a WorksapceWindow it will determine whether the window is running and then
+ * call `createChild` or `positionWindow` as appropriate.
+ */
```
ryanpetersonOF: Spelling in this whole comment block

tomrobertsOF: Oh wow... I must have had a brain aneurism when typing that. I did a quick spelling pass before making the PR but apparently missed lots of them.

tomrobertsOF:
@pjbroadbent, think I've addressed everything we discussed yesterday. Should be ready for a second look.

src/client/restoreHelpers.ts
tomrobertsOF: Should all be resolved now.

src/client/restoreHelpers.ts
```diff
@@ -0,0 +1,125 @@
...
+        defaultLeft: bounds.left,
+        defaultTop: bounds.top,
+        frame,
+        state,
+        name
+    });
+}
+
+/**
+ * Given a workspaceWindow object for a currently open window, this will position and
```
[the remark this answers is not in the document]
tomrobertsOF: That was just a typo. Fixing now.

src/client/restoreHelpers.ts
```diff
@@ -0,0 +1,125 @@
+import {WorkspaceWindow, WorkspaceApp} from './workspaces';
+
+/**
```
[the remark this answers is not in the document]
tomrobertsOF: 👍  on the `@link`. 

For the `@throws` we really are just rethrowing the runtime errors. Not sure if there's a sensible way of documenting that.

src/client/restoreHelpers.ts
```diff
@@ -0,0 +1,125 @@
+import {WorkspaceWindow, WorkspaceApp} from './workspaces';
```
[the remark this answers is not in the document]
tomrobertsOF: 👍

src/client/restoreHelpers.ts
```diff
@@ -0,0 +1,134 @@
+/**
+ * @module RestoreHelpers
+ */
+import {WorkspaceWindow, WorkspaceApp} from './workspaces';
+
+/**
+ * A simple restore handler that can be passed to {@link setRestoreHandler} when no custom logic is needed.
+ *
+ * This restore handler will open and position any child windows which are not currently running
```
bryangaleOF: Would change to "This restore handler will open any child windows which are not currently running, and move new and existing child windows to their expected positions for the workspace being restored." (Avoiding using both 'move' and 'position' as verbs, and making a call that a workspace gets restored rather than it being the workspace that does the restoring)

src/client/restoreHelpers.ts
```diff
@@ -0,0 +1,134 @@
...
+ * and move any open child windows to their expected positions for the restoring workspace.
+ *
+ * ``` ts
+ * import {workspaces, restoreHelpers} from 'openfin-layouts';
+ *
+ * workspaces.setRestoreHandler(restoreHelpers.standardRestoreHandler);
+ * ```
+ *
+ * @param workspaceApp Will contain details of the running app's child windows and their expected post-restore state.
```
bryangaleOF: Consider if we need 'details' here. Could it be "Will contain the expected post-restore state of the running app's child windows"?

tomrobertsOF: My thinking was the details are in the Type definitions and in the comments about restoring in general in the `Workspaces` module

src/client/restoreHelpers.ts
tomrobertsOF: Like that. Will change

tomrobertsOF:
@bryangaleOF, just pushed up a couple more small tweaks based on your comments.

src/client/restoreHelpers.ts
tomrobertsOF: I added a quick addition saying where to look for more information on what this means

PR #418, opened 2019-07-17
[SERVICE-528] Prevent windows from being left off-screen after (dis-)connecting a monitor

bryangaleOF:
This PR:
- Adds `MonitorAssignmentCalculator`, which handles the maths of figuring out the new position of windows
- Adds `MonitorAssignmentValidator`, which handles actually moving windows around
- Adds call to `MonitorAssignmentValidator` when a monitors changed, which is debounced similarly to tab and snap group validation
- Integration tests for overall behaviour, and unit tests specifically around `MonitorAssignmentCalculator`


src/provider/snapanddock/utils/Debounced.ts
```diff
@@ -47,27 +57,31 @@ export class Debounced<C extends Function, S, A extends any[]> {  // tslint:disa
...
+        await this._callback.apply(this._scope, args);
 
-        this.handle = -1;
-        this.callback.apply(this.scope, args);
+        const resolve = this._deferredPromise![1];
+        this._deferredPromise = undefined;
+        resolve();
     }
 
     private cancel(): void {
```
tomrobertsOF: Would it make sense for `cancel` to reject the deferred promise?

bryangaleOF: This was used to cancel the timeout immediately before it was being restarted, so in this case we still want the promise to resolve successfully.

`cancel()` was only called once, and should be clearer now that I've inlined it

tomrobertsOF: Ok, makes sense. 👍

src/provider/model/MonitorAssignmentCalculator.ts
```diff
@@ -0,0 +1,119 @@
...
+type DesktopSnapGroup = {entities: DesktopEntity[]} & Rectangle;
+
+export class MonitorAssignmentCalculator {
+    private _monitorRectangles: ReadonlyArray<Rectangle>;
+
+    public constructor(monitorRectangles: ReadonlyArray<Rectangle>) {
+        this._monitorRectangles = monitorRectangles;
+    }
+
+    public getMovedSnapGroupRectangles(snapGroup: DesktopSnapGroup): SnapGroupResult {
```
pjbroadbent: From a naming perspective, is there a better verb than "get"?

Like `resolveSnapGroup`/`resolveEntity`? Or maybe not resolve, but validate/calculate/find<X>Position?

src/provider/model/MonitorAssignmentCalculator.ts
```diff
@@ -0,0 +1,119 @@
...
+        return {
+            entityRectangle: this.attemptGetEntityRectangleInsideMonitorRectangle(stateEntityRectangle, primaryMonitor).rectangle,
+            monitorRectangle: primaryMonitor
+        };
+    }
+
+    /**
+     * For a given entity rectangle and monitor rectangle, returns if possible, a new rectangle for the entity that fits entirely within the monitor
+     */
+    private attemptGetEntityRectangleInsideMonitorRectangle(entityRectangle: Rectangle, monitorRectangle: Rectangle): {rectangle: Rectangle, inside: boolean} {
```
pjbroadbent: Descriptive, but a bit verbose?..

`fitEntityWithinBounds`?..

src/provider/model/MonitorAssignmentCalculator.ts
```diff
@@ -0,0 +1,119 @@
...
+                // If the window doesn't fit, make a best-effort to display it sensibly
+                inside = false;
+
+                if (axis === 'y') {
+                    // In the y axis case, position window so title bar is at top of monitor
+                    resultRectangle.center[axis] = entityRectangle.halfSize.y + monitorRectangle.center.y - monitorRectangle.halfSize.y;
+                } else {
+                    resultRectangle.center[axis] = monitorRectangle.center[axis];
+                }
+            } else {
```
pjbroadbent: Would this be equivalent to what's below, without the separate upper and lower handling?

```
const offset = (entityRectangle.center[axis] - monitorRectangle.center[axis]);

if (Math.abs(offset) > buffer) {
    // Window is (partially) off-screen. Move inward to touch the monitor edge.
    resultRectangle.center[axis] = monitorRectangle.center[axis] + (Math.sign(offset) * buffer);
}
```

src/provider/model/MonitorAssignmentValidator.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private getSnapGroups(): DesktopSnapGroup[] {
+        return this._model.snapGroups.filter(snapGroup => snapGroup.isNonTrivial());
+    }
+
+    private getEntities(): DesktopEntity[] {
+        const trivalEntities = this._model.snapGroups.filter(snapGroup => !snapGroup.isNonTrivial());
+        // Equivalent of trivalEntities.flatMap()
+        return Array.prototype.concat(...trivalEntities.map(snapGroup => snapGroup.entities));
```
pjbroadbent: Can you avoid the `concat` here?

```
return trivialEntities.map(snapGroup => snapGroup.entities[0]);
```

src/provider/model/MonitorAssignmentValidator.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private async applySnapGroupResults(snapGroups: DesktopSnapGroup[], snapGroupResults: SnapGroupResult[]): Promise<void> {
```
pjbroadbent: Would it make sense for `SnapGroupResult` to contain a reference to the "source" group? Rather than having parallel arrays?

src/provider/model/MonitorAssignmentValidator.ts
```diff
@@ -0,0 +1,114 @@
...
+        await Promise.all(snapGroupResults.map(async (snapGroupResult, snapGroupIndex : number) => {
+            const snapGroup = snapGroups[snapGroupIndex];
+
+            await this.applySnapGroupResult(snapGroup, snapGroupResult.groupRectangle);
+            await this.applyEntityResults(snapGroup.entities, snapGroupResult.entityResults);
```
pjbroadbent: Slightly confused by this, wouldn't this apply the offsets twice?

Or could the group-level move "corrupt" the entity-level move?

src/provider/model/MonitorAssignmentValidator.ts
```diff
@@ -0,0 +1,114 @@
...
+        if (offset.x !== 0 || offset.y !== 0) {
+            if (entity.snapGroup.isNonTrivial()) {
+                await entity.setSnapGroup(new DesktopSnapGroup());
+            }
+
+            const oldState = entity instanceof DesktopTabGroup ? entity.state : entity.currentState.state;
+
+            // Things get weird with tabs if we mess with them while minimized, and we'll want to re-maximize maximized windows
+            // to the correct screen, so always restore
+            if (oldState !== 'normal') {
```
pjbroadbent: Do we also need to restore if the tab group is maximized? A maximized tab group actually consists of all "normal" windows, so there shouldn't be any "hidden state" to mess us up later?

src/provider/model/DesktopEntity.ts
```diff
@@ -90,4 +98,25 @@ export interface DesktopEntity {
...
+     */
+    minimize(): Promise<void>;
+
+    /**
+     * Maximizes the entity
+     */
+    maximize(): Promise<void>;
+
+    /**
+     * Restores the entity. Following this, the window may be maximized or normal
```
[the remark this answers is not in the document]
bryangaleOF: Not if it's applied to a window that's been minimized then maximzed

src/provider/snapanddock/utils/Debounced.ts
```diff
@@ -8,21 +10,23 @@
...
-    private scope: S;
-    private args: A|undefined;
+    private _deferredPromise: DeferredPromise<void> | undefined;
+    private _result: any;
+
+    private _callback: C;
+    private _scope: S;
+    private _args: A|undefined;
 
     // Multiple definitions of setTimeout/clearTimeout, and not possible to point TSC at the correct (non-Node) definition
```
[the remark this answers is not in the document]
bryangaleOF: ✅

src/provider/snapanddock/utils/Debounced.ts
```diff
@@ -47,27 +62,32 @@ export class Debounced<C extends Function, S, A extends any[]> {  // tslint:disa
...
+    private async schedule(elapsed: number = 0): Promise<void> {
+        if (this._handle !== undefined) {
+            clearTimeout(this._handle as number);
         }
-    }
 
-    private schedule(): void {
-        this.cancel();
-        this.handle = setTimeout(this.onTimeout, Debounced.DEBOUNCE_INTERVAL);
+        this._handle = setTimeout(this.onTimeout, Math.max(0, Debounced.DEBOUNCE_INTERVAL - elapsed));
```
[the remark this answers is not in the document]
bryangaleOF: The `elapsed` here preserves existing behaviour. Otherwise for `async` calls, we'd be increasing the wait

src/provider/model/DesktopTabGroup.ts
```diff
@@ -288,33 +315,26 @@ export class DesktopTabGroup implements DesktopEntity {
...
-            this.window.sendEvent(event);
+                const bounds: Rectangle = this._normalTabBounds;
+                await this._window.applyProperties({
+                    center: {x: bounds.center.x, y: bounds.center.y - bounds.halfSize.y - (this._config.height / 2)},
+                    halfSize: {x: bounds.halfSize.x, y: this._config.height / 2}
+                });
+                await this.activeTab.applyProperties(bounds);
+            }
+        } else {
+            await Promise.all(this._tabs.map(tab => tab.applyProperties({state: 'normal'})));
```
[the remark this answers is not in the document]
bryangaleOF: This is a "this line was here and I'm scared to touch it" thing

src/provider/model/DesktopWindow.ts
```diff
@@ -693,6 +693,19 @@ export class DesktopWindow implements DesktopEntity {
...
+    public async maximize(): Promise<void> {
+        return this.applyProperties({state: 'maximized'});
+    }
+
+    public async minimize(): Promise<void> {
+        return this.applyProperties({state: 'minimized'});
+    }
+
+    public async restore(): Promise<void> {
+        // Note that the actual end state following this may be 'maximized'
```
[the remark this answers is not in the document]
bryangaleOF: It's the 'minimized maximized' window case again. Our internal state will still be correct as we listen for maximized events and update the state accordingly

src/provider/model/MonitorAssignmentCalculator.ts
bryangaleOF: I think `calculate` if anything, but would argue redundant from the class name

src/provider/model/MonitorAssignmentCalculator.ts
bryangaleOF: `fit` implies side-effects to me. Private method for something very specific, I feel ok with descriptive-at-the-cost-of-clunky

src/provider/model/MonitorAssignmentValidator.ts
bryangaleOF: That wouldn't be the same, and without bumping the Typescript version, I think this is the best way we have. Fundamentally, we ware `concat`ing a bunch of arrays - we're turning [[A, B]. [C, D]] into [A, B, C, D]

src/provider/model/MonitorAssignmentValidator.ts
bryangaleOF: In theory no, and we could expose a new method on TabGroup to change the normalBounds and remaxize to whatever its new monitor is to be, but this is for expediency

src/provider/model/MonitorAssignmentValidator.ts
bryangaleOF: Hmm, would feel odd to have getMovedSnapGroupRectangles return its input param. Could bundle the two together in the map in line 33?

bryangaleOF: Oh, but then applyEntityResults will end up following a different pattern. Unless...

We'd need to have both `SnapGroupResult` and `EntityResult` containing references to the input param, which, ... ehhhh

src/provider/model/MonitorAssignmentValidator.ts
bryangaleOF: Nah, as what we get back from the calculator is absolute rectangles, not offsets

src/provider/model/MonitorAssignmentCalculator.ts
bryangaleOF: *squints*

Yes.

bryangaleOF: ✅

src/provider/model/DesktopTabGroup.ts
```diff
@@ -193,6 +207,19 @@ export class DesktopTabGroup implements DesktopEntity {
     public async applyOffset(offset: Point, halfSize?: Point): Promise<void> {
         const tabstripHalfHeight: number = this._config.height / 2;
 
+        // Emulating DesktopWindow behaviour, restore the window in the maximized case
+        if (this._isMaximized) {
+            if (this._normalTabBounds) {
+                // The given halfSize is for the group, but we store the halfSize for a tab, so we do the below conversion
```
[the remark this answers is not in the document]
bryangaleOF: Yes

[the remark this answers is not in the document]
bryangaleOF: ✅

src/provider/snapanddock/utils/Debounced.ts
```diff
@@ -47,27 +62,32 @@ export class Debounced<C extends Function, S, A extends any[]> {  // tslint:disa
...
+        this._handle = undefined;
+        const resolve = this._deferredPromise![1];
+        this._deferredPromise = undefined;
+
+        this._result = this._callback.apply(this._scope, args);
+        await this._result;
 
-        this.handle = -1;
-        this.callback.apply(this.scope, args);
+        this._result = undefined;
```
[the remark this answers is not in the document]
bryangaleOF: The difference between `_deferredPromise` and _result is that _result doesn't get setup until the call is in progress. We wouldn't want to be awaiting on `_deferredPromise` in the initial `call()` method.

But yeah, this whole thing is kinda tricky to reason about, should go through with you

src/provider/model/DesktopEntity.ts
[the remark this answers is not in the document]
pjbroadbent: `restore` will maximise the window in those cases then?

Does that apply to tab groups too? And does that mean that our state will be incorrect (`normal`, when it should be `maximized`) or does that get picked up and corrected through some event listeners?

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "compilerOptions": {
         "outDir": "dist",
         "rootDir": "src",
-        "types": ["node"]
+        "types": ["node"],
```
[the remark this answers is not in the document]
bryangaleOF: ✅

test/demo/monitorChange.inttest.ts
```diff
@@ -0,0 +1,354 @@
...
+            await window.minimize();
+        }
+    }
+}
+
+async function setupSnapAndTabGroups(context: WindowContext, testOptions: MonitorAssignmentValidatorTestOptions) {
+    await promiseForEach(testOptions.initialGrouping.snap, async (snapGroup) => {
+        const window = context.windows[0];
+
+        for (let i = 1; i < context.windows.length; i++) {
```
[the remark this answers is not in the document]
bryangaleOF: Neat!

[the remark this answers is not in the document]
bryangaleOF: ✅

src/provider/model/DesktopEntity.ts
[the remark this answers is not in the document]
pjbroadbent: Saw the other reply. 👍

src/provider/utils/async.ts
```diff
@@ -0,0 +1,15 @@
+export type DeferredPromise<T> = [Promise<T>, (value?: T) => void, (reason?: any) => void]
+
+/**
+ * Creates a deferred promise and returns it along with handlers to resolve/reject it imperatively
+ * @returns a tuple with the promise and its resolve/reject handlers
+ */
+export function deferredPromise<T = void>(): DeferredPromise<T> {
```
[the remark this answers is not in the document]
bryangaleOF: Actually, may as well just do it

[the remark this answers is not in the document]
bryangaleOF: ✅

src/provider/model/MonitorAssignmentValidator.ts
pjbroadbent: Am I missing something here? We're filtering-out only the trivial groups, so there would always only be one entity within the array?

Protecting against corrupt/empty groups?..

src/provider/model/MonitorAssignmentValidator.ts
pjbroadbent: Yeah, was thinking we'd keep both `SnapGroupResult` and `EntityResult` consistent.

In my mind, makes sense for an `XResult` to bundle both "X" and what X maps to. Means that those results can then be handled atomically - kinda like `SnapTarget`/`TabTarget`/etc.

Minor point though, if objections can leave as-is.

src/provider/model/MonitorAssignmentValidator.ts
pjbroadbent: But those absolute positions get applied as offsets (well, which are then mapped to absolute positions again)... Does "feel" wrong, but seems like it would work.

Still does seems like there's some redundancy here. The flow is a little hard to follow.

src/provider/snapanddock/utils/Debounced.ts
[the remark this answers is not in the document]
pjbroadbent: Missed some of the nuance in how the elapsed was calculated. Makes sense.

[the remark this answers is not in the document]
bryangaleOF: Committed improved naming and commenting

src/provider/snapanddock/utils/Debounced.ts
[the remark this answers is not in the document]
bryangaleOF: Committed improved naming and commenting

src/provider/model/MonitorAssignmentValidator.ts
pjbroadbent: As discussed, leave as-is. The use of simplified types makes the above a little more tricky.

Ultimate solution may well be some tooling/utils. Having workable mocks for snap group (etc) would help with this. Kinda tricky in Layouts, though.

bryangaleOF: Gonna leave. Could do something smart with generics, but it's a non-urgent rabbit hole

src/provider/model/MonitorAssignmentValidator.ts
bryangaleOF: Changed

✅

src/provider/model/MonitorAssignmentValidator.ts
bryangaleOF: Commented as discussed

✅

src/provider/model/DesktopTabGroup.ts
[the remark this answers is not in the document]
bryangaleOF: Tracing through each path, the only difference I'm seeing is that in the minimized case we now await on `applyProperties` (which is actually instantaneous anyway) before sending the `tab-group-restored` event

src/provider/model/MonitorAssignmentCalculator.ts
bryangaleOF: Shortened name

✅

src/provider/model/MonitorAssignmentCalculator.ts
bryangaleOF: Where did we leave this one? Acceptable to leave as is?

pjbroadbent: Yep

PR #427, opened 2019-08-02
[SERVICE-565] Hardening of monitor change edge-cases

bryangaleOF:
This PR improves handling of monitor disconnection/resizing:
- We now take advantage of the fact that windows composing a DesktopTabGroup are never maximized (in the runtime sense), to reduce the amount of faff needed
- We also take advantage of the fact we can move 'minimized' windows, but can't rely on runtime group behaviour
- We properly handle the "minimized maximized window case
- We now check for overlapping entities in group validation. This is also now extracted into its own file, and unit tested
- We always use `availableRect` rather than `monitorRect`. Use was inconsistent before


src/provider/model/DesktopTabGroup.ts
```diff
@@ -849,17 +859,36 @@ export class DesktopTabGroup implements DesktopEntity {
...
-                }
+            await DesktopWindow.transaction([this._window, ...tabsToReposition], async () => {
+                await this._window.applyProperties(expectedTabstripPosition);
+            });
+        }
 
-                for (const tab of tabsToReposition) {
-                    await tab.applyProperties(expectedTabPosition);
-                }
+        if (tabsToReposition.length > 0) {
```
bryangaleOF: This change to now use two distinct transactions to fix-up TabGroups was prompted by an issue where the DesktopSnapGroups at the end of the transaction could be out of sync with runtime-level groups. Now that I've split this into two transactions, this doesn't occur.

Bug with transactions? Or was I using it wrong before?

src/provider/utils/groups.ts
```diff
@@ -0,0 +1,93 @@
+import {Rectangle, RectUtils} from '../snapanddock/utils/RectUtils';
+import {MIN_OVERLAP, ADJACENCY_FUZZ_DISTANCE} from '../snapanddock/Constants';
+
+type DesktopEntity<T> = {currentState: Rectangle & {hidden: boolean}} & T;
+
+export function getContiguousEntities<T = {}>(entities: DesktopEntity<T>[]): T[][] {
+    const contiguousSets: DesktopEntity<T>[][] = [];
+    const disjointEntities: DesktopEntity<T>[] = [];
+
+    // Extract any overlapping entities into trivial groups
```
bryangaleOF: This isn't optimal for minimizing the number of groups we split our invalid group into, but it does avoid needing to make arbitrary choices about which group to place an overlapping window in, and any entity order dependence that may follow from that

tomrobertsOF: I'm in favour of this approach. Any choice we make would likely seem arbitrary on use. This way there is no ambiguity.

tomrobertsOF: Also, having now properly read the code it looks like we will ungroup any window with any overlap. I feel like we could maybe be a bit smarter and only ungroup windows that overlap with other windows in their group. 

For examples right now we have:
 - A and B are in a group and don't overlap
 - C (not in a group) overlaps with B
 - B and C are both moved to trivial groups
and
 - A, B, and C are in a group
 - B and C overlap
 - B and C are both ungrouped.
I think the second situation is fine, but the first situation should not ungroup any windows.

bryangaleOF: In practice the entities passed into `getContiguousEntities` are all of the same group. Could this just be a naming/commenting change?

tomrobertsOF: Oh yea, of course. I forgot where this was called. Comment tweaks may help other people looking at this down the line, but not sure exactly what to change. I'll resolve this for now, but feel free to make some adjustments if you think of something.

src/provider/model/DesktopTabGroup.ts
```diff
@@ -355,6 +302,21 @@ export class DesktopTabGroup implements DesktopEntity {
         this.window.sendEvent(event);
     }
 
+    public get isMaximized(): boolean {
```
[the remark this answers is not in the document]
bryangaleOF: ✅

src/provider/model/DesktopSnapGroup.ts
```diff
@@ -218,12 +217,19 @@ export class DesktopSnapGroup {
     }
 
     public async applyOffset(offset: Point): Promise<void> {
-        return this.rootWindow!.applyOffset(offset);
+        if (this.rootWindow!.currentState.state === 'minimized') {
+            // Windows don't move as a group when minimized, so in this case move all windows independently
+            return DesktopWindow.transaction(this.windows, async (windows) => {
```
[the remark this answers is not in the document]
bryangaleOF: Yep 👍

Makes me nervous that I've missed a case where minimized windows do group, but will give this change some testing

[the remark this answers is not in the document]
bryangaleOF: Hmm, doesn't look good from testing. Have updated the comment here - "Windows can't be relied on to move as a group when minimized"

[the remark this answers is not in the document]
tomrobertsOF: Ok. If it doesn't work that's fine, just want to make sure we really need the transactions we use.

src/provider/utils/groups.ts
bryangaleOF: Added block comment. Think function name works pretty well so have left as is

tomrobertsOF: Fine by me. This is your comment chain so I won't mark it as resolved.

src/provider/utils/groups.ts
```diff
@@ -0,0 +1,98 @@
...
+
+type DesktopEntity<T> = {currentState: Rectangle & {hidden: boolean}} & T;
+
+/**
+ * Takes a list of entities and returns the groups these entities form, based on their size and position. Each array of entities
+ * returned will form a valid group (no entities overlap, and all entities are transitively adjacent). Broadly tries to minimize
+ * the number of arrays returned, and if possible will return a single array containing all input entities.
+ */
+export function getContiguousEntities<T = {}>(entities: DesktopEntity<T>[]): T[][] {
+    const contiguousSets: DesktopEntity<T>[][] = [];
```
TheoOttahOF: Minor detail but this is easier to notice, plus you use this type later on to create a new array.
`Array<DesktopEntity<T>[]>`

src/provider/utils/groups.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    while (unvisited.length > 0) {
+        const visited: DesktopEntity<T>[] = [];
+        depthFirstSearch(unvisited[0], visited);
+        contiguousSets.push(visited);
+    }
+
+    return contiguousSets;
+
+    function depthFirstSearch(startEntity: DesktopEntity<T>, visited: DesktopEntity<T>[]) {
```
TheoOttahOF: Can this be turned into a pure function that returns the contiguous sets? 
Feels weird that a search function alters external variables instead of returning the result based on its input. Also removes the need of the while loop.

src/provider/utils/groups.ts
bryangaleOF: We're generally not using Array<>, and I'd rather not use it where we don't need to or mix the two types of declaration. I've gone the other direction and modified how the adjacency list is built

src/provider/utils/groups.ts
bryangaleOF: I don't think you can get rid of any of the loops without fundamentally changing the approach. Two loops to extract a contiguous set (breadth-wise and depth-wise traversal of the adjacency table, which we've got with the `for` loop and the recursive call), and then to keep doing that until we've got all the contiguous sets (the `while` loop).

And with how we've got our adjacency data, we need some way to track what's already been searched. That _could_ move outside the recursive function, but it gives more bookkeeping to do outside of that, which I think would be messier.

Admittedly the nested function is weird in the context of our codebase, but it's driven by having this recursive approach, which is natural for doing a depth first search.

I've had a go at tidying it, and getting rid of the visited and unvisited lists. Is this an improvement?

So I'd change the while loop to:

    // Find all contiguous sets
    while (disjointEntities.length > 0) {
        contiguousSets.push(extractFirstAdjacentSet(disjointEntities, adjacencyLists));
    }

get rid of the nested function, and add a new function:

    function extractFirstAdjacentSet<T>(disjointEntities: DesktopEntity<T>[], adjacencyLists: DesktopEntity<T>[][]): DesktopEntity<T>[]
    {
        const accumulator: DesktopEntity<T>[] = [];

        recursiveDepthFirstSearch(disjointEntities[0]);

        return accumulator;

        function recursiveDepthFirstSearch(entity: DesktopEntity<T>): void {
            const index = disjointEntities.indexOf(entity);

            if (index === -1) {
                return;
            } else {
                const adjacencyList = adjacencyLists[index];

                disjointEntities.splice(index, 1);
                adjacencyLists.splice(index, 1);

                accumulator.push(entity);

                for (const adjacentEntity of adjacencyList) {
                    recursiveDepthFirstSearch(adjacentEntity);
                }
            }
        }
    }

bryangaleOF: Pushed that change, but open to reverting if we don't feel like it's an improvement

src/provider/utils/groups.ts
tomrobertsOF: I'm with Bryan on this. We don't use `Array<...>` unless it is an absolute requirement.

src/provider/utils/groups.ts
TheoOttahOF: This methods nicer. As you said its legacy code so not worth spending more time on it.

PR #428, opened 2019-08-06
[SERVICE-539] Validate window positions after workspace restore

bryangaleOF:
This PR:
- Adds a preprocessing step to the restore process that fixes up a workspace object to fit within the current monitor arrangement
- Adds some funky generics stuff to `MonitorAssignmentCalculator`, to remove the need for parallel arrays from `MonitorAssignmentValidator`, and to make its use in `retargetWorkspaceForMonitors` easier
- Makes a bunch of changes so to not drag `main.ts` into our unit test (most of the file changes are down to this, so the file count makes this PR look bigger than it really is)

Changes to flag
- `retargetWorkspaceForMonitors` mutates the workspace directly, which is something I'd usually avoid, but eels preferable to creating a deep copy of this large object
- The workspace returned to the client is the *modified* workspace, rather than the original. Is this fine? What d we expect clients to do with the returned workspace anyway?
- To save a bunch of `as WindowIdentity` casts, I've modified the types in `src/client/workspaces.ts`. This shouuuuuuld be fine, ...right?  Like, clients shouldn't ever be writing to this field directly? Or is this more dangerous than I think?


pjbroadbent:
> The workspace returned to the client is the modified workspace, rather than the original. Is this fine? What d we expect clients to do with the returned workspace anyway?

I think this was the intention of the return value. Been a while since I've checked, but I think the service already returns a modified version of the workspace - in the event that some windows fail to restore, etc.

src/provider/workspaces/monitor.ts
```diff
@@ -0,0 +1,322 @@
...
+            for (const window of entity.windows) {
+                window.bounds.left += offset.x;
+                window.bounds.right += offset.x;
+                window.bounds.top += offset.y;
+                window.bounds.bottom += offset.y;
+            }
+        }
+    }
+
+    private getRectangleFromWindow(window: WorkspaceWindow): Rectangle {
```
pjbroadbent: This duplicates existing RectUtil stuff - this could make use of `convertToCenterHalfSize`, or could remove this and just use that helper directly?

As in `this.getRectangleFromWindow(window)` => `RectUtils.convertToCenterHalfSize(window.bounds)`

src/provider/workspaces/monitor.ts
```diff
@@ -0,0 +1,322 @@
...
+            halfSize: {x: window.bounds.width / 2, y: window.bounds.height / 2}
+        };
+    }
+
+    private getRectangleFromTabGroup(tabGroup: TabGroup): Rectangle {
+        const tabstripConfig = tabGroup.groupInfo.config;
+        const tabstripHeight = tabstripConfig === 'default' ? DEFAULT_TABSTRIP_HEIGHT : tabstripConfig.height;
+
+        const height = tabGroup.groupInfo.dimensions.appHeight + tabstripHeight;
+        const dimensions = tabGroup.groupInfo.dimensions;
```
pjbroadbent: Can we have types on these vars?

src/provider/workspaces/restore.ts
```diff
@@ -96,11 +102,15 @@ export const restoreWorkspace = async(payload: Workspace): Promise<Workspace> =>
         group.validate();
     }
 
-    const event: WorkspaceRestoredEvent = {type: 'workspace-restored', workspace};
-    apiHandler.sendToAll(EVENT_CHANNEL_TOPIC, event);
+    if (haveMonitorsBeenDetached(monitors, model.monitors)) {
```
pjbroadbent: How robust is this? It's relying on an implementation detail of `DesktopModel`. Should we clone the initial monitors array above, to ensure that this won't ever be an "if `x == x`" check?

test/provider/retargetWorkspaceForMonitors.unittest.ts
```diff
@@ -0,0 +1,707 @@
...
+                left: 250,
+                right: 550,
+                top: 300,
+                bottom: 500,
+                width: 300,
+                height: 200
+            }
+        }
+    }]
+};
```
pjbroadbent: This is getting a bit out-of-hand. We need to bump SERVICE-517 up the priority list, get some utils that avoid all this really verbose data within the tests themselves.

src/client/workspaces.ts
```diff
@@ -162,7 +162,7 @@ export interface WorkspaceWindow extends WindowIdentity {
      *
      * These groupings will be restored once all windows in the workspace have been re-created.
      */
-    windowGroup: Identity[];
+    windowGroup: WindowIdentity[];
```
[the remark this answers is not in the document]
bryangaleOF: Looking at where and how this gets populated (from the result of `getGroup` in `workspaces/group.ts`), it would be crashing if the identities didn't already have both uuid and name

src/provider/model/MonitorAssignmentCalculator.ts
```diff
@@ -1,22 +1,22 @@
...
+export type EntityResult<T> = {rectangle: Rectangle, target: T};
+export type SnapGroupResult<T, U> = {entityResults: EntityResult<T>[], groupRectangle: Rectangle, target: U};
 
 type EntityAndMonitorRectangles = {entityRectangle: Rectangle, monitorRectangle: Rectangle}
 
 // Limit ourselves to just the bits of DesktopEntity and SnapGroups we want
-type DesktopEntity = {normalBounds: Rectangle};
-type DesktopSnapGroup = {entities: DesktopEntity[]} & Rectangle;
+type DesktopEntity<T> = {normalBounds: Rectangle} & T;
+type DesktopSnapGroup<T, U> = {entities: DesktopEntity<T>[]} & Rectangle & U;
```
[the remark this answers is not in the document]
bryangaleOF: Done

src/provider/utils/identity.ts
```diff
@@ -0,0 +1,15 @@
+import {Identity} from 'hadouken-js-adapter';
```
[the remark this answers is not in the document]
bryangaleOF: Done

src/provider/workspaces/monitor.ts
```diff
@@ -0,0 +1,322 @@
...
+
+                if (groupedEntity) {
+                    groupedEntities.push(groupedEntity);
+                }
+            }
+            groupedEntities.push(entity);
+
+            groupedEntities.sort((entity1, entity2) => {
+                const id1 = entity1.type === 'window' ? getId(entity1.window) : getId(entity1.activeTab);
+                const id2 = entity2.type === 'window' ? getId(entity2.window) : getId(entity2.activeTab);
```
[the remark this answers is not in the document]
bryangaleOF: We _could_ do that. But the `id` in our fake `DesktopEntity` would mean something different (id of the window or the active tab) to what it means in the model (id of the window or the tabstrip), which feels like a tripwire to be setting someone

src/provider/workspaces/monitor.ts
bryangaleOF: Why these two in particular? There's lots of these function local variables without explicit types, where the name strongly suggests what type it is. Or do you want types throughout?

test/provider/retargetWorkspaceForMonitors.unittest.ts
bryangaleOF: Thoughts on me pulling out the data into its own file for now?

src/provider/workspaces/monitor.ts
```diff
@@ -0,0 +1,322 @@
...
+import {haveMonitorsBeenDetached} from '../utils/monitor';
+import {DEFAULT_TABSTRIP_HEIGHT} from '../utils/constants';
+import {getId} from '../utils/identity';
+
+type WindowPartialEntity = {type: 'window', window: WorkspaceWindow};
+type TabGroupPartialEntity = {type: 'tabGroup', tabGroup: TabGroup, windows: WorkspaceWindow[], activeTab: WorkspaceWindow};
+type Entity = (WindowPartialEntity | TabGroupPartialEntity) & {normalBounds: Rectangle};
+type SnapGroup = Rectangle & {entities: Entity[]};
+
+export function retargetWorkspaceForMonitors(workspace: Workspace, monitors: ReadonlyArray<Rectangle>): void {
```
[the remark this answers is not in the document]
bryangaleOF: Done

src/provider/utils/monitor.ts
```diff
@@ -0,0 +1,5 @@
+import {Rectangle, RectUtils} from '../snapanddock/utils/RectUtils';
+
+export function haveMonitorsBeenDetached(oldMonitors: ReadonlyArray<Rectangle>, newMonitors: ReadonlyArray<Rectangle>): boolean {
```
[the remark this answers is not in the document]
bryangaleOF: Changed to method on Model

src/provider/workspaces/restore.ts
bryangaleOF: Done

src/provider/workspaces/monitor.ts
```diff
@@ -0,0 +1,322 @@
...
+        const tabGroupsById = new Map<string, TabGroup>();
+        for (const tabGroup of this._workspace.tabGroups) {
+            for (const tab of tabGroup.tabs) {
+                tabGroupsById.set(getId(tab), tabGroup);
+            }
+        }
+
+        // Create entities lookup, with window identity, or active tab identity as key
+        const entitiesById = new Map<string, Entity>();
+        for (const window of this._windowsById.values()) {
```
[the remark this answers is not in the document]
bryangaleOF: Don't _think_ so. We do lookups within `this._windowsById` inside here, so we want `this._windowsById` to alreadybe fully populated

src/provider/workspaces/monitor.ts
bryangaleOF: Done

bryangaleOF:
@pjbroadbent All code changes responded to. If all's get let talk about the schema thing

src/provider/workspaces/monitor.ts
pjbroadbent: Sorry, was referring to throughout. Comment is here since these were the first I spotted.

We're typically fairly consistent about putting types on variables, so was more for consistency than anything else. Minor point.

test/provider/retargetWorkspaceForMonitors.unittest.ts
pjbroadbent: As an intermediate step between embedding in the code and having template support, could move each of these to a JSON file (which we'd later change into a template file).

As in move the mock workspace into a `res/test/mocks/workspace.json` file, then `require()` that file within the test/util.

src/provider/workspaces/monitor.ts
bryangaleOF: On these kind of very local, short lived things, I don't agree we are, and it's more of a case-by-case thing...

bryangaleOF:
Removed `WorkspaceWindow` change, fixed bug that meant monitor validation only happened when monitors *hadn't* changed (introduced in last round of post-feedback changes), and condensed down test data a bit.

src/provider/model/DesktopModel.ts
```diff
@@ -153,13 +154,13 @@ export class DesktopModel {
      * @param identity Window identifier - either a UUID/name object, or a stringified identity as created by @see getId
      */
     public getWindow(identity: WindowIdentity|string): DesktopWindow|null {
-        const id = typeof identity === 'string' ? identity : this.getId(identity);
+        const id = typeof identity === 'string' ? identity : getId(identity);
```
ryanpetersonOF: Super petty but its included elsewhere - theres no type on this.  This is the case on a few others as well

PR #431, opened 2019-08-23
[RUN-5434]/[RUN-5435] Change restore to use startManyManifests

MichaelMCoates:
This PR uses `startManyManifests` to start a batch of applications, instead of running them individually. This avoids the issue reported earlier of applications failing to come up during a large restore.

This PR is linked to the following PRs:
https://github.com/HadoukenIO/core/pull/907
https://github.com/HadoukenIO/js-adapter/pull/325
https://github.com/openfin/openfinrvm/pull/307

### DIFF
--- a/src/provider/workspaces/restore.ts
+++ b/src/provider/workspaces/restore.ts
@@ -61,10 +61,14 @@ export const restoreWorkspace = async(payload: Workspace): Promise<Workspace> =>
 
     const workspace = payload;
     const startupApps: Promise<WorkspaceApp>[] = [];
+    const manifestApps: WorkspaceApp[] = [];
 
     await createWorkspacePlaceholders(workspace);
 
-    const apps: WorkspaceApp[] = await promiseMap(workspace.apps, app => restoreApp(app, startupApps));
+    const apps: WorkspaceApp[] = await promiseMap(workspace.apps, app => restoreApp(app, startupApps, manifestApps));
+
+    //@ts-ignore
+    fin.Application.startManyManifests(manifestApps);
 
     // Wait for all apps to start-up
     const startupResponses: WorkspaceApp[] = await Promise.all(startupApps);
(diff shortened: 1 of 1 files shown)


src/provider/workspaces/restore.ts
```diff
@@ -255,26 +258,17 @@ const restoreApp = async(app: WorkspaceApp, startupApps: Promise<WorkspaceApp>[]
...
                     console.error('Unable to restart programmatically launched app:', app);
                 }
             }
-
-            if (ofAppNotRunning) {
-                // Application.run() can hang with createFromManifest calls, so we set a timeout to
-                // continue restoration even if run() hangs.
-                // We throw an exception to continue restoration, so take a look at the functions in our
-                // catch to see how we handle it.
-                await attemptToRunCreatedApp(ofAppNotRunning);
```
pjbroadbent: Seems this function is now unused - remove?

src/provider/workspaces/restore.ts
```diff
@@ -255,26 +258,17 @@ const restoreApp = async(app: WorkspaceApp, startupApps: Promise<WorkspaceApp>[]
...
-                await model.expect({uuid, name: uuid});
-                await positionWindow(app.mainWindow, true);
```
pjbroadbent: Where does the positioning of the main window happen in this flow?

I only see `positionWindow` calls for main windows of apps that are already running.

src/provider/workspaces/restore.ts
```diff
@@ -61,10 +61,14 @@ export const restoreWorkspace = async(payload: Workspace): Promise<Workspace> =>
...
     const startupApps: Promise<WorkspaceApp>[] = [];
+    const manifestApps: WorkspaceApp[] = [];
 
     await createWorkspacePlaceholders(workspace);
 
-    const apps: WorkspaceApp[] = await promiseMap(workspace.apps, app => restoreApp(app, startupApps));
+    const apps: WorkspaceApp[] = await promiseMap(workspace.apps, app => restoreApp(app, startupApps, manifestApps));
+
+    //@ts-ignore
+    fin.Application.startManyManifests(manifestApps);
```
pjbroadbent: What's the plan for managing dependencies here? Assuming this won't be merged in just yet?

Will the updated types be available before this is merged, to avoid the `ts-ignore`?

src/provider/workspaces/restore.ts
MichaelMCoates: Correct, removing.

src/provider/workspaces/restore.ts
MichaelMCoates: In this flow, positioning is handled in the placeholder windows: 
https://github.com/HadoukenIO/layouts-service/blob/develop/src/provider/workspaces/placeholder.ts#L51

I think that this double-positioning was a leftover from when restoration was significantly more unstable.

src/provider/workspaces/restore.ts
MichaelMCoates: Yes! We're planning to update the types along with the .44 release, so this will be removed once that's in.

